PR #49, opened 2015-10-21
Addrhash

cdkersey:
Finish support for address hashing in memHierarchy cache. Reduces set contention which greatly improves performance on some applications. Max latency for HARP drops from 20k cycles to 200 cycles using the XOR-based hash function.


cdkersey:
Oops-- this clearly should have gone to devel.

PR #71, opened 2016-01-17
Fix Zodiac DUMPI Compile Errors

nmhamster:
Removes compile time errors in the changed interface name for Hermes. This is exposed when DUMPI is enabled for Zodiac.


nmhamster:
Error - should not be to master. Delete.

PR #89, opened 2016-01-26
Afrodri/fast siev

afrodri:
This is a branch to support a "Fast" memory sieve. Instead of a large histogram, it tracks actual malloc()s and free()s. 

Modifications: 
  – To ariel: add new links so ariel cores can pass allocation / free events back to the memsieve
  – To ariel front-end: add pass back the IP address before a malloc
  – to sieve: new links to get alloc events from ariel, and new tracking to match misses to allocations and print out stats at the end.


nmhamster:
Is this tested with and without the memsieve being attached? This is a pretty heavy change to Ariel so I'd like to understand how we can integrate that into testing if possible?

afrodri:
I haven't tested it without memsieve, only with memsieve but without using the change (i.e. turning off the allocation tracking, not making the extra allocation links). 
Are there any pre-canned ariel tests in the elements repo, or are they in the SQE repo?

jpvandy:
There are six Ariel test input files in this directory:

sst/elements/ariel/frontend/simple/examples/stream

afrodri:
I've tested with those (Except for memHstream.py). They seem to run to completion without errors (some warnings about stats changes). Are there sample "correct" outputs to compare against?

Issue #152, opened 2016-03-14
SST Patterns should be moved to the Elements Attic


nmhamster:
Move Patterns into the legacy elements attic. No longer supported as of SST v6.0.0.

nmhamster:
This issue is going to hold up the external-elements work because there is an m4 macro in patterns for detecting where MPI is enabled by SST. I need to make sure we get this out ASAP.

allevin:
Pull Request 156 corrects the issue.

PR #156, opened 2016-03-15
Moved patterns to archive per issue #152

allevin:
Simply removed the patterns directory. The same directory was added to sst-elements-archive repo.


allevin:
Auto merge will fail, due to failures in testing.  This is a high priority change, so merge is done manually

PR #159, opened 2016-03-15
Afrodri/circ

afrodri:
Circuit Counter network inspector - counts the number of circuits (unique source/dest pairs) that pass through a router.


afrodri:
Closing pull request to resubmit with inspector subdir

PR #184, opened 2016-03-25
fix a condition where the firefly NIC misses data on a VC

mjleven:


nmhamster:
Set to master and not devel. Reopened on PR#185.

PR #185, opened 2016-03-25
Fix a condition where the firefly NIC misses data on a VC

nmhamster:


nmhamster:
Issue with Github and AutoTester connectivity. We are overriding this to perform a merge since all tests have successfully passed.

PR #193, opened 2016-03-29
update Firefly to use new Params interface

mjleven:


nmhamster:
Closing because this is merge to master, should have been devel.

PR #194, opened 2016-03-29
Change FireFly to use the new Params Interface for SST v6.0

nmhamster:
Change FireFly to use the new Params Interface for SST v6.0


nmhamster:
Tests have passed by AutoTesrer is having issues so overriding while this gets fixed.

PR #196, opened 2016-03-30
Enable Prospero Profiling from an SST Command Line Tool

nmhamster:
Enables Prospero tracing to be taken from an SST support command line tool which decides how to run the underlying PIN process. This tool is successfully able to launch PIN on Linux and Mac OSX.


nmhamster:
Argh .. this is on master. Closing.

PR #201, opened 2016-03-31
Provide a Tool to Generate Prospero Memory Traces

nmhamster:
Provides a tool to correctly load PIN and perform a multi-threaded memory trace for Prospero. Appears that PIN detection and execution in testing for Prospero is not consistent with how Ariel is used even though both tools use PIN. This means Prospero will fail but Ariel will not. This tool provides the same launch mechanism as Ariel but for Prospero traces. Support is provided for `text`, `binary` and `compressed` traces.

### DIFF
--- a/prospero/Makefile.am
+++ b/prospero/Makefile.am
@@ -74,7 +74,7 @@ all-local: tracetool/sstmemtrace.cc
 	-L$(PINTOOL_DIR)/extras/xed-intel64/lib \
 	-o prosperotrace.dylib tracetool/sstmemtrace.cc \
 	-stdlib=libstdc++ \
-	-lpin -lxed -lpindwarf -lpthread $(LIBZ_LDFLAGS)
+	-lpin -lxed -lpindwarf -lpthread $(LIBZ_LDFLAGS) $(LIBZ_LIB)
 	
 install-exec-hook: prosperotrace.dylib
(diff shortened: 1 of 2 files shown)


jpvandy:
@nmhamster  Ariel PIN tests were running sst-test only.  Prospero Pin test were running NO where.   Your new version allow the Prospero Pin test to run on CentOS.
The appropriate PIN modules are also installed on Ubuntu-14.04 and Yosemite and El Capitan.

Trying essentially the same test Suite on Ubuntu and Yosemite produce very different failures.

On Yosemite the reported prospero.pin failure is:
----
Child process(es) for profiling completed.
===============================================================
 pin return flag is from compressed: 0
E: Application address overlaps with Pin, try to relink application /Users/sstbuild/jenkins/workspace/SST__XXX_memH_no_openMP_Yosemite-Prospero/44_2016-04-29_13-43-15/devel/trunk/sst/elements/prospero/tests/array/array: Cannot allocate memory
===============================================================
-----------
On Yosemite the Prospero Pin Refusal is:

E: Attach to pid 18443 failed. 
E:   The Operating System configuration prevents Pin from using the default (parent) injection mode.
E:   To resolve this, either execute the following (as root):
E:   $ echo 0 > /proc/sys/kernel/yama/ptrace_scope
E:   Or use the "-injection child" option.
E:   For more information, regarding child injection, see Injection section in the Pin User Manual.
E:

PR #203, opened 2016-04-03
Fix miranda gups generator seeds not set problem

shavvn:
there're *seed_a* and *seed_b* in the params, but they're never set, so it's not actually random...


nmhamster:
@shavvn thank you for the patch!

shavvn:
@nmhamster  np :)

PR #221, opened 2016-04-11
Merge from Master into Devel to Propagate Arun's PR

nmhamster:
Last week we accidentally had an automatic merge of a PR into `master`. This request performs a merger of those changes back into `devel`.


nmhamster:
Re-try this as a merge into devel with PR #223. Close this one. May need manual merge fixing.

Issue #218, opened 2016-04-11
Reorganize Ariel/Makefile.am PIN Compilation


nmhamster:

nmhamster:
Opened new PR under #235

nmhamster:
Merged.

PR #222, opened 2016-04-12
Modify Ariel Makefile.am to use Automake install hooks

nmhamster:
Modify Ariel Makefile.am to use Automake install hooks. Address issue #218.


nmhamster:
Needs to be fixed on Linux. Close and will re-open.

PR #224, opened 2016-04-12
Allow Miranda::RandomGen to be seeded from input deck

nmhamster:
Allow Miranda::RandomGen to be seeded from input deck using the new `seed` parameter and re-seeding capability of `SST::RNG`.


nmhamster:
Needs to be fixed according to build output. Close for now.

PR #230, opened 2016-04-13
Modify Ariel Makefile.am to use Automake install hooks

nmhamster:
Modify Ariel Makefile.am to use Automake install hooks and match the equivalent in Prospero.


nmhamster:
Close and attempt to re-open.

PR #305, opened 2016-05-17
Mjleven/miranda

mjleven:
added support for other components to drive Miranda


nmhamster:
Against master, will not be merged.

PR #339, opened 2016-05-31
Attempt to pick up -mt suffix on boost libraries

bjmoor:
Attempt to automatically determine of the -mt suffix on BOOST libraries is needed, and if so, have Scheduler component use the suffix.


jwilso:
Jenkins is disallowing jobs from starting in order to reboot to fix an issue.  This is normal; jobs will resume after reboot.

--
Jon Wilson
SST Software Engineering/Quality Assurance
CSRI
Sandia National Laboratories

Issue #424, opened 2016-07-13
Compilation fails on Archlinux: scheduler missing include


luisfnqoliveira:
I get a compilation error on Archlinux (and other related errors):
`DragonflyMachine.cc:122:78: error: ‘ceil’ was not declared in this scope`

on the following file:
`sst-elements/src/sst/elements/scheduler/DragonflyMachine.cc`
`sst-elements/src/sst/elements/scheduler/StencilMachine.cc`
`sst-elements/src/sst/elements/scheduler/Torus3DMachine.cc`

The fix is easy, in those files add:
`#include <cmath>`

This error does not occur in Ubuntu 14.04.

OS: Archlinux
GCC:6.1.1
Boost:1.56.0
OpenMPI:1.8
SST-Core commit: eb8280abb3d535dff3813f909d0b8bc67010669a
SST-Elements commit: 5c0e78b0510b273c9cfcfd078b0e43e0a524caf1

allevin:
This issue was corrected with PR #425

PR #425, opened 2016-07-13
Solved a compilation issue in Archlinux

luisfnqoliveira:
Solution to issue #424.
In archlinux the compiler did not find the math library functions. Added an include in files with issue.


allevin:
SST is preparing for 6.0 Release, autotesting resources are limited for a while, testing for this PR may fail a few times, but will eventually be run properly.

nmhamster:
@luisfnqoliveira thanks for the fix!

PR #428, opened 2016-08-08
Fix ember bugs

shavvn:
I was looking at the code in /ember/run/ but wasn't able to run it due to various bugs there... Here's my fix and I'm wondering if this is a deprecated folder since there's a lot of similar code in /ember/test/ that works.


nmhamster:
Thanks @shavvn for the bug fixes!

PR #479, opened 2016-10-21
Memory controller cleanup

mjleven:
changed the interface between the memory controller and the backends


afrodri:
I think I understand these changes, but what is the goal of the change? I.e. is it really cleaner pass around more arguments rather than a struct?

PR #550, opened 2017-01-06
Dynlink

h4u5:
This is the start of an extension to merlin which adds dynamic link width capabilities. By default this is disabled. When enabled, if a port is idle below some threshold (dlink_thresh) the port adjust its output_timing to half (representing a reduction in link width. Currently this is checked every 10 microseconds by PortControl::handleSAIWindow.

When a link width is reduced, the port is prevented from sending messages for a hardcoded amount of time (1microsecond). The code that disables the port-sends are two "if statements" inserted into handle_output_r2r and handle_output_r2n. The port is then reenabled in PortControl::reenablePort.
The code inserted into handle_output functions should be double checked to make sure it doesn't break any existing functionality when dynamic link widths are enabled.

I've updated the constructors for portControl and to calls made within hr_router to include an optional dlink_thresh parameter. This parameters specifies whether links may dynamically reduce or increase link width. A value of dlink_thresh=-1 disables the functionality (default). A value between 0 and 1 specifies the required proportion of time that a link must be idle before reducing its link width.
Instructions for Issuing a Pull Request to sst-elements.

Future enhancements would be:

    A more sophisticated way to determine when to disable a port.
    Exposing the interval to handleSAIWindow to the user.
    Determining whether to count disabled time as idle or stalled in terms of SAI (stalled/active/idle)
    Additional link width adjustments beyond decreasing by half or increasing by two. For example for a 12X link we may want to allow for 6X, 3X and 1X widths.
    Tying this into a power model automatically.


allevin:
Approve Pull Request App - Pull Request Inspected & Approved
allevin Has Inspected and Approved code changes to Pull Request
Comment From allevin = Code inspected and approved by S. Hammond

allevin:
Approve Pull Request App - Pull Request Inspected & Approved
allevin Has Inspected and Approved code changes to Pull Request
Comment From allevin = Fork updated from devel branch to allow testing to continue.

PR #551, opened 2017-01-06
Cramsim print cmd trace

mbhealy:
Added a simple addition for printing DRAM command traces either to stdout or a specified file. Controllable using config file parameters.


nmhamster:
@allevin PR is approved for testing to start.

PR #632, opened 2017-04-11
Add initial testcase for HDF5 statoutput

bjmoor:
Add a test-case for HDF5-based statoutput.   @jpvandy Can you enable running the `sdl9-2-hdf.py` and comparing its textual output against the reference file, and the HDF5 file against the '`out.h5`' file?   This test can only run on systems where SST-Core found HDF5.


bjmoor:
Closing this PR temporarily until an SST-Core PR finishes migration.

Issue #536, opened 2016-12-08
Weekly Valgrind test no longer runs clean on Ariel and memHierarcy-sdl tests


jpvandy:
Valgrind finds unitialize variable use in the Ariel Ivy and Sandy Bridge tests  and also in the memHierarchy sdl tests.

Ivy bridge Valgrind output file is attached.
[Valgrind-ivb.txt](https://github.com/sstsimulator/sst-elements/files/640825/Valgrind-ivb.txt)

Here is the output for memH sdl-1
[Valgrid_sdl_1.txt](https://github.com/sstsimulator/sst-elements/files/640833/Valgrid_sdl_1.txt)

sst-elements version is:  54d2d87cfd37553a3c99867514e0bf6f4c04a7f3

bjmoor:
@jpvandy With PR #633, is this issue closable?

PR #633, opened 2017-04-12
Fix uninitialized variable

bjmoor:
Valgrind was reporting use of uninitialized variable in some statistics.  Tracked down to this variable being uninitialized.   In C++, POD typed member variables are not initialized to zero.


bjmoor:
This may address #536

PR #637, opened 2017-04-17
Add check for boost to configure

bjmoor:
Now that BOOST is optional for core, ELEMENTS needs its own check for boost.


bjmoor:
How is it not an error?  Multiple SST-ELEMENT libraries require boost to compile.   Until #595 is done, we don't have the option for cherry-picking which libraries to build.

nmhamster:
@bjmoor - I thought we were making it so that SST elements which don't depend on Boost will still be compiled. Its an optional dependency for most of them. This change makes it an enforced dependency.

bjmoor:
Fair enough.  I will work on that.   A quick grep on my part suggests just MemH and CramSim will require it.

nmhamster:
@bjmoor I'm a bit confused why MemH will require it, would be good to understand if its a hard requirement or it would just disable a small piece of functionality.

bjmoor:
MemH uses it because it uses boost::variant.  Basically, a type-safe Union construct.   std::variant  only comes along in C++17.

Could it be refactored out?  Probably.  But not today.

bjmoor:
@nmhamster Please re-review

PR #774, opened 2017-09-15
backend logic for mapping existing memory requests to any other request type in goblinHMCbackend

jleidel:
Backend logic for mapping existing memory requests to any other request type in goblinHMCbackend.  This also includes the necessary parameter logic to map CMC command structures as the backing request type for those commands.  This set of commits also includes a new ExtMemBackendConvertor that successfully converts the memory requests for all existing types to what is now necessary in the ExtMemBackend types.  

Note: The find_array parameter passing is currently broken on the SST Core Master TOT.  I'm sending this pull request now as the included patches have tests that can be used to reproduce the issue.  

---

----


jleidel:
Note, find_array parsing is not broken; but it does need to be a bit more study.

allevin:
Approve Pull Request App - Pull Request Inspected & Approved
allevin Has Inspected and Approved code changes to Pull Request
Comment From allevin = Code reviewed and approved by S. Hammond

PR #794, opened 2017-10-18
Ship memLinkBase.h as a public header

KarateSnowMachine:
memNIC.h is a public header that includes memLinkBase.h and so
memLinkBase.h should also be shipped.


allevin:
@KarateSnowMachine Testing will continue every 4 hours with a fail result.  I am going to close the PR.  When you have updated your fork/branch to match the latest from sst-elements/devel, you can reopen the PR.

PR #838, opened 2017-12-01
fixing HBMDramSim test fixtures to reflect new subcomponent name

jleidel:
fixing HBMDramSim test fixtures to reflect new subcomponent name

This fixes the currently failing tests for HBMDRAMSim2.  The only change here is to enforce the new naming convention when loading the subcomponent through memHierarchy.


allevin:
@jleidel @gvoskuilen This is not failing our nightly tests.  Is this some sort of test that you are running separately with your own testbench? or are we missing something in our test coverage?  The code that was merged into master was living in the devel branch for a few days.

gvoskuilen:
@allevin These are tests that @jleidel runs. We aren't currently testing HBMDramSim but we should start testing it before the 8.0 release.

PR #914, opened 2018-02-19
Adding Stake Miranda subcomponent that fuses RISC-V Spike simulator to SST for memory simulation

jleidel:
Adding the Stake "generator" subcomponent to Miranda.  This subcomponent provides a shim layer for the RISC-V "Spike" simulator to drive the SST memory simulation components.  All the standard functional simulation layers in the RISC-V Spike (riscv-isa-sim) infrastructure remain.  Stake provides a simple generator with a C-based callback routine that allows Spike to directly send memory requests to SST.  With these commits, we can directly drive memory experiments with RISC-V ELF binaries (including booting Linux on Spike)

Caveats: 
- This requires a modified riscv-isa-sim that is currently located here: https://github.com/tactcomplabs/riscv-isa-sim/tree/sst
- Atomic and custom memory operations are currently mapped to read requests.  We will update this in future commits to provide more flexible configuration of atomics
- We did not commit any tests with this pull request.  Spike executes ELF binary payloads.  As a result, we elected not to commit individual binary blobs into the SST source tree.  If the main SST developers feel that this is ok, we can most certainly provide sample execution scripts.

### DIFF
--- a/config/sst_check_stake.m4
+++ b/config/sst_check_stake.m4
@@ -1,0 +1,49 @@
+AC_DEFUN([SST_CHECK_STAKE],
+[
+  sst_check_stake_happy="yes"
+
+  AC_ARG_WITH([stake],
+    [AS_HELP_STRING([--with-stake@<:@=DIR@:>@],
+      [Use RISC-V Spike simulator installed in optionally specified DIR])])
+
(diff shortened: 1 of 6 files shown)


jleidel:
Can I get a failure log of the AutoTester such that I can triage the issue?

jleidel:
build failed due to including "stake.h" when not enabled; working on a fix now

PR #929, opened 2018-02-28
Disagg amro

amroawad2:
Erase this information and put your Pull Request comments here 
---

1 - Verify that the Pull Request is targeted to the **devel** branch of sstsimulator/sst-elements

2 - Verify that Source branch is up to date with the devel branch of sst-elements

3 - After submitting your Pull Request:
   * Automatic Testing will commence in a short while 
      * Pull Requests will be tested with the devel branches of the sst-core and sst-sqe repositories
         * These branches are syncronized with the devel branch of sst-elements.  This is why is it important to keep your source branch up to date.
      * If testing passes, the source branch will be automatically merged (if possible)
         * Pull Requests from forks will not be automatically tested until the code is inspected.
         * Pull Requests from forks will not be automatically merged into the devel branch.
      * If testing fails, You will be notified of the test results.  
         * The Pull Request will be retested on a regular basis - Changes to the source branch can be made to correct problems
         
4 - DO NOT DELETE THE BRANCH (OR FORKED REPO) UNTIL THE PULL REQUEST IS MERGED.
----

### DIFF
--- a/src/sst/elements/Opal/Opal.cc
+++ b/src/sst/elements/Opal/Opal.cc
@@ -40,7 +40,7 @@ Opal::Opal(SST::ComponentId_t id, SST::Params& params): Component(id) {
  registerAsPrimaryComponent();
 //    primaryComponentDoNotEndSim();
 
-    int verbosity = (uint32_t) params.find<uint32_t>("verbose", 1);
+    verbosity = (uint32_t) params.find<uint32_t>("verbose", 1);
 	output = new SST::Output("OpalComponent[@f:@l:@p] ", verbosity, 0, SST::Output::STDOUT);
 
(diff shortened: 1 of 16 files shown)


gvoskuilen:
@amroawad2 @VamseeReddyK Does the 'node' parameter need to be known by caches/memories/etc. or just the memNIC? If only the memNIC needs it, you could simplify by adding it as a parameter for memNIC (or if you need it for direct links - 'memLink' too, memLinkBase) instead of everywhere.

VamseeReddyK:
@gvoskuilen, node parameter should be known by memLinkBase.h. As you said it can be simplified by adding it as parameter to memNIC and memLink. But then I thought it would be difficult to configure by the user. I have used this parameter only for 3 components. For caches, directly controller and memory controller. If you feel this is not required I can simplify it.

PR #939, opened 2018-03-08
fixed bug in ariel core by updating clock cycles only during valid cl…

VamseeReddyK:
Ariel core clock should be updated only when coreQ->empty() or refillQueue() succeeds, at-least once in a clock tick.

### DIFF
--- a/src/sst/elements/ariel/arielcore.cc
+++ b/src/sst/elements/ariel/arielcore.cc
@@ -615,6 +615,8 @@ bool ArielCore::processNextEvent() {
 		}
 	}
 
+	updateCycle = true;
+
 	ARIEL_CORE_VERBOSE(8, output->verbose(CALL_INFO, 8, 0, "Processing next event in core %" PRIu32 "...\n", coreID));
 
 	ArielEvent* nextEvent = coreQ->front();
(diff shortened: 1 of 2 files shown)


nmhamster:
@VamseeReddyK - I'm not sure about this change. The code in question needs to count cycles if the Ariel clock is incremented otherwise it won't have correct timing. The `processEvent` function may return false because all units are busy in which case it is correct to increment the clock.

VamseeReddyK:
@nmhamster This is the case where the speed at which the pin tool (application at test, user application) injects instructions to ariel core queue is less than ariel core tick speed. In this case a lot of ariel core ticks do nothing, just because the core queue is empty and refill queue fails, but the clock cycles counter is updated. 

If core queue is not empty or refill queue is successful at-least once during multiple processEvent() calls in ariel core tick function's 'for' loop, the counter should be updated. 

If core queue is empty then ariel core tries to refill the queue. If nothing is available to refill the queue then processEvent() would return false but counter is not updated only if the processEvent() is called for the first time during ariel clock tick. If not the counter is updated as an event is processed in the previous processEvent() call during ariel core tick.

nmhamster:
@VamseeReddyK - so I don't think this is the correct fix. Basically, the problem is that if SST is ticking the Ariel core then time is progressing anyway. SST controls the timing at the PDES core and other components will have moved forward. If Ariel does not get enough events, we need to work out how to fix this issue, not adjust the statistics. One option is that we could introduce some blocking for Ariel to prevent further execution without events. We should talk about your setup because this is extremely rare.

allevin:
@VamseeReddyK @nmhamster The autotester improperly merged this PR due to internal technical reasons.  @VamseeReddyK you did nothing wrong.  

@nmhamster and I are discussing how to correct the Autotester and will be making changes to prevent this in the future.  

However, this PR needs to be reverted (Si, you are probably better at this than I - Can you do it?).  Be aware that a number of PR's have been merged into Devel since this PR was merged.

PR #964, opened 2018-04-02
convert to new ELI

mjleven:


allevin:
Turning on the WIP label until the autotester is back online.

allevin:
@mjleven The Autotester is back online, These failures are real.

PR #965, opened 2018-04-03
convert Ember to new ELI

mjleven:


allevin:
Turning on the WIP label until the autotester is back online.

PR #990, opened 2018-04-23
Fixed a typo in an array index that was causing memory corruption.

feldergast:


feldergast:
Closing because the branch wasn't merged with devel pre-PR.

PR #991, opened 2018-04-23
fix Firefly NIC send machine 

mjleven:
The send machine logic gets in a state where it wasn't parallelizing sends.


allevin:
@mjleven The testing system has stabilized.  I believe that these are real failures.

PR #1000, opened 2018-04-28
TLBShootdown: Initial commit. Works for one node

VamseeReddyK:


VamseeReddyK:
@nmhamster Yes, Ariel works as expected when Opal is not enabled. All my changes will be executed only when opal is enabled.

src/sst/elements/ariel/arielcore.h
```diff
@@ -52,6 +52,7 @@
 
 #include "ariel_shmem.h"
 #include "arieltracegen.h"
+#include <../Opal/Opal_Event.h>
```
gvoskuilen: Should be "sst/elements/Opal/Opal_Event.h"

VamseeReddyK: @gvoskuilen Should I close this PR, change the code and create a new PR?

gvoskuilen: @VamseeReddyK no need to close the PR, just make the change and push it to your fork

src/sst/elements/ariel/arielcore.cc
```diff
@@ -218,6 +219,35 @@ void ArielCore::handleEvent(SimpleMem::Request* event) {
 	delete event;
 }
 
+void ArielCore::ISR_Opal(OpalEvent *ev) {
```
gvoskuilen: @VamseeReddyK Can you pull the latest version of devel and merge it into your fork/branch? It has a new method for stalling ariel cores which you should be able to use instead of setting/clearing maxIssuePerCycle. See the 'isStalled' variable.

VamseeReddyK: @gvoskuilen I have some more commits in my branch for which I want to create a PR. Do I need to close this PR, merge disagg_amro branch with devel and then create a new PR?

PR #1013, opened 2018-05-05
Disagg amro

VamseeReddyK:


hughes-c:
This isn't merging with devel, only with the disagg_amro branch.

allevin:
Actually the sstsimulator:disagg_amro branch may be removed very soon (post SST 8) as we move  to the forked repos workflow.  This will leave us with just a few branches for master, devel, and the release branches.  Feature branches will be done in peoples forks.

PR #1029, opened 2018-05-11
fix condition where a body packet is processed before the hdr delay

mjleven:


allevin:
@mjleven Is your branch up to date with devel?  Scott was having problems with this the DetailedNetwork test, and fixed it a few days ago.

PR #1030, opened 2018-05-12
TLB shootdown and Page migration

VamseeReddyK:


src/sst/elements/Samba/PageTableWalker.cc
```diff
@@ -308,7 +395,7 @@ void PageTableWalker::recvResp(SST::Event * event)
 
 	insert_way(WID_Add[pw_id], find_victim_way(WID_Add[pw_id], WSR_COUNT[pw_id]), WSR_COUNT[pw_id]);
 
-	long long int addr = WID_Add[pw_id];
```
hughes-c: Not required but it would be good to have an address type to make the code more readable. Something like` typedef uint64_t Address_t`

src/sst/elements/Samba/PageTableWalker.cc
```diff
@@ -279,17 +312,71 @@ void PageTableWalker::handleEvent( SST::Event* e )
...
+		*shootdown = 1;
+		shootdownId = ev->getShootdownId();
+
+		uint64_t newPaddress = ev->getPaddress();
+		uint64_t vaddress = ev->getAddress();
+		int faultLevel = ev->getFaultLevel();
+
+		// update physical address. get the level to update
+		// only by core 0
+		if(id == 0) {
```
hughes-c: Try to follow http://sst-simulator.org/SSTPages/SSTDeveloperCodingStandards

hughes-c:
- test_Messierxxxx failes because test files were changed to exclude min/max values --> should remove updated messier test files
- test_Sambaxxxx failes because test files were changed to exclude min/max values --> should correct
[out.txt](https://github.com/sstsimulator/sst-elements/files/2017642/out.txt)

allevin:
@VamseeReddyK @amroawad2 @hughes-c

The Opal directory has a file called src/sst/elements/Opal/! (yes ! is the file name).  Is this correct?  if Not, can I delete it (I will do this in a separate PR)?

src/sst/elements/ariel/arielcore.cc
```diff
@@ -901,7 +901,7 @@ void ArielCore::tick() {
         currentCycles++;
         statCycles->addData(1);
 
-        if( updateCycle ) {
+        if( updateCycle || isStalled) {
```
hughes-c: Why do you need to update activeCycles if the core is stalled?

VamseeReddyK: For example if the core is waiting for the completion of shootdown event, the core is stalled but the cycle needs to be counted as cycles are wasted in such scenarios. So cycles need to updated while the core is stalled.

hughes-c: Current cycles is updated, which corresponds to the stat cycles. statActiveCycles corresponds to the stat active_cycles. Active cycles are only updated when events are processed -- the stall flag prevents new events from being issued.

VamseeReddyK: @hughes-c Can I use a new stat variable(stalled_cycles) which just counts the number of cycles for which the core is stalled?

PR #1045, opened 2018-05-23
Fix bug in API

hughes-c:


nmhamster:
@hughes-c - can you explain what issue this fixes (which bug)? Should you include a ".c" file in a ".h" file?

hughes-c:
@nmhamster 
The api libraries would not link without including the function definitions via the .c in the header. 
gcc  -L../../api someProg.c -o someProg.x -larielapi

Failed because it could not find ariel_enable() even though it was clearly present in both the static and dynamic library.

static
0000000000000000 T ariel_enable
0000000000000010 T ariel_fence
                 U puts

dynamic
0000000000000695 T ariel_enable
00000000000006a7 T ariel_fence

PR #1107, opened 2018-09-06
Pymerlin update

feldergast:
Added new python modules sst.merlin.core, sst.merlin,endpoint and sst.merlin.topology.  This is still beta and not all topologies and endpoints have been implemented.


allevin:
@feldergast is this PR still valid?  It is 5 months old.

PR #1136, opened 2018-10-02
Generic Neural Architecture, Initial merge

afrodri:
This is an implementation of a 'Generic Neural Architecture' based on the STPU work (Hill et All, ICRC 2017, "A Spike-Timing Neuromorphic Architecture", 10.1109/ICRC.2017.8123631). 

It is a largely self-contained component, though it uses the simpleMem interface. 
There is a test and reference file in sst-elements/src/sst/elements/GNA/tests


allevin:
@afrodri Make sure your branch is up to date with devel.

PR #1156, opened 2018-10-29
Mjleven/reorg ember load

mjleven:
moved EmberEP from loadInfo.py to EmberEP.py
move load from file logic to emberLoad
other general cleanup


hughes-c:
Waiting on fix for jenkins slaves (I think).

jpvandy:
Attempting to switch MAC testing from El Capitan to High Sierra

PR #1241, opened 2019-03-14
Allow SimpleMem to use the max_requests_per_cycle parameter from MemBackend

hughes-c:


hughes-c:
Need to update tests to use  "max_requests_per_cycle": "1" or gold reference files

PR #1247, opened 2019-03-29
Shogun

hughes-c:


nmhamster:
@hughes-c - I think we are good to do the first merge if you are happy?

hughes-c:
May need to tweak the tests and input_message_slots doesn't do anything in the de-clocking version. Up to you whether or not we want to leave that variable in. I would like to go back and implement it, I just haven't had time.

PR #1259, opened 2019-04-17
Reference file updates for testing

hughes-c:


hughes-c:
Shogun tests not comparing with refFiles...

Issue #1253, opened 2019-04-03
Shogun Option input_message_slots


hughes-c:
The input_message_slots configuration option has no affect other than changing the number of events that the crossbar will propagate each cycle via the moveEvents() arbiter function.

hughes-c:
Fixed in https://github.com/sstsimulator/sst-elements/pull/1266

PR #1263, opened 2019-04-23
Bug Fix for Issue 253

hughes-c:
This fixes the bug described in issue https://github.com/sstsimulator/sst-elements/issues/1253. Because there's no way to apply back pressure, it adds a hidden queue on the input port but only allows in_msg_per_cycle to propagate to the actual port queue

This also updates the test environment.

- Added new test for multi-message crossbar

- Updated test files to reflect changes in paremeters
&nbsp;&nbsp;&nbsp;input_message_slots -> in_msg_per_cycle
&nbsp;&nbsp;&nbsp;output_message_slots -> out_msg_per_cycle

- Updated test result reference  files to reflect limits on inputs messages


hughes-c:
Might be better to just modify the num_events that each port will process each cycle. The side effect of this is that the queues may fill.

PR #1301, opened 2019-06-24
Updated SingleXStream

hughes-c:
Help for SingleStream was incorrect; it listed the parameters start_at and stop_at but the subcomponent expects startat and stopat.

Updated ReverseSingleStream to be consistent with other subcomponents, using startat and stopat.

This is a fix for https://github.com/sstsimulator/sst-elements/issues/1302


jpvandy:
Clay,

Mojave (and hence autotester) is down since Friday/Saturday.   Jon came in, from being home sick, to work on it.

No ETA at the moment.

   John

PR #1307, opened 2019-07-01
Mygensa

jpvandy:
pull the gensa tests to devel


jpvandy:
close for retry after clean up of last week's Ariel problem

PR #1364, opened 2019-09-12
Added GPGPU-Sim configure option

hughes-c:
New configure flag: --with-gpgpusim

Balar also updated to accept 	GPGPUSIM_LIBDIR and  GPGPUSIM_LIB, which should point to the modified cudart library, cudart_mod.


hughes-c:
The libtool m4 file uses awk, so I'm leaning toward leaving it as is.

PR #1377, opened 2019-10-02
Updating For Balar

hughes-c:
Updating paths and classes for name change.


hughes-c:
This isn't going to pass until SQE has been updated with https://github.com/sstsimulator/sst-sqe/pull/1002

Issue #1417, opened 2020-01-23
--with-sst-core=corepath can find incorrect sst-core if in $PATH


allevin:
1 - Detailed description of problem or enhancement

During configure, --with-sst-core=corepath will improperly find an previously built core if it is in the $PATH rather than the user defined "corepath".

Basically if a previously built core in the the $PATH, the --with-sst-core will find that one rather than the corepath defined by the user.  The configure script will report which one it found, but if the user does not double check, they may believe that the configure worked and selected the corepath they wanted.  This can cause a newer SST-Elements to be built with an older SST-core (that might be defined in the path).

2 - Describe how to reproduce
- Build & Install SST-Core in directory "PATH_A" and add that dir "PATH_A" to the $PATH.  
- Build & Install SST_Core in directory "PATH_B", and do not add to the $PATH.
- Configure SST-Elements with --with-sst-core="B".  The configure will find $PATH_A" not the user desired "PATH_B"

3 - What Operating system(s) and versions 
ALL

4 - What version of external libraries (Boost, MPI)
N/A

5 - Provide sha1 of all relevant sst repositories (sst-core, sst-elements, etc)
Affects the current sst-elements devel branch from sha 8d36dc and earlier.

6 - Fill out Labels, Milestones, and Assignee fields as best possible

allevin:
Reported by @deanchester 

Hi Aaron, 

I stripped back the configure scripts (the previous ones had been used to build version 8) and CramSIM still fails to build with the same error. I also deleted the clones of the repositories and recloned them from Github and ran the autogen scripts again. 

I am trying to build on OS X 10.13.6

When I reviewed the configuration output its pulling SST core from my path/environment and not from the one specified to configure. 
-------------------------------------------------------
SST Elements Configuration completed.
-------------------------------------------------------
Build Environment:

                  SST-Elements Version : -dev
                            Git Branch : master
                          Git HEAD SHA : ad60d81078a9a1dca0ffc04411858491953c6a73
                   Branch Commit Count : 11926
                   SST-Elements Prefix : /Users/deangchester/local/sstelements-9
                       SST-Core Prefix : /Users/deangchester/local/sstcore
                          Preprocessor : gcc -E
                          C++ Compiler : g++
                            C Compiler : gcc
                      Compiler Version : Apple LLVM version 10.0.0 (clang-1000.11.45.5)
[...]

PR #1418, opened 2020-01-23
Changes to fix sst-core detect if another core is in the path

allevin:
This PR is to address issue #1417.

During configure, --with-sst-core=corepath will improperly find an previously built core if it is in the $PATH rather than the user defined "corepath".

Basically if a previously built core in the the $PATH, the --with-sst-core will find that one rather than the corepath defined by the user. The configure script will report which one it found, but if the user does not double check, they may believe that the configure worked and selected the corepath they wanted. This can cause a newer SST-Elements to be built with an older SST-core (that might be defined in the path).

There are basically 3 changes:
1 - Lines 10 and 19 fix incorrect logic.  This test was always failing because $with_sst_core actually returns the user defined path, not a "yes".  This caused lines 12 and 21 (fail condition) to always execute.
2 - Lines 12 and 21 would search the $PATH first for a copy of sst-core then if not found it would search the user defined path.  The change here basically will force a check **only** of the user defined path if the --with-sst-core is defined.  if --with-sst-core is not defined, the script will search $PATH for a copy of sst-core.
3 - Slight change to the error message on lines 17 and 26 showing preference for the user defined directory.

With this change, if core is not found in the user defined path the configure will fail.  It will no longer attempt to search the $PATH and possibly find a different core that the user did not want.  

However if --with-sst-core is not provided, then the script will search $PATH for a core, and if not found will cause configure to fail.

@gvoskuilen @feldergast can you please review and if/when approved, I will remove the WIP flags and allow the autotester to run and merge this PR.

### DIFF
--- a/config/sst_core_check_install.m4
+++ b/config/sst_core_check_install.m4
@@ -2,28 +2,32 @@
 AC_DEFUN([SST_CORE_CHECK_INSTALL], [
 	AC_ARG_WITH([sst-core],
 	  [AS_HELP_STRING([--with-sst-core=@<:@=DIR@:>@],
-	    [Use SST Discrete Event Core installed in DIR])])
+	    [Use SST Discrete Event Core installed in DIR])],
+	    [with_sst_core=$withval/bin],
+	    [with_sst_core=$PATH])
 
   SST_CONFIG_TOOL=""
   SST_REGISTER_TOOL=""
 
-  AS_IF( [test "x$with_sst_core" = "x"],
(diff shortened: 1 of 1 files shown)


allevin:
@gvoskuilen @feldergast 
I implemented Gwen's suggestion, and I like this version much more.  It detected if --with-sst-core is defined without a path (and also defined without an equal sign).  If --with-sst-core is not defined, it will only search the path, and if it is defined, it will search the user defined directory.  

Please review and then I will release for auto-testing.

PR #1425, opened 2020-01-29
Implement a generic method to mark an element as deprecated.

allevin:
Implement a generic method to mark an element as deprecated.  

This is a similar technique adding the .ignore into the element directory to have it ignored during autogen/configure.

we just add a hidden file called `.deprecated` and the configure script will detect it and mark the the element as YES - DEPRECATED in the configuration summary.

The real code changes happen on line 179 (my editor decided to clean up the extra whitespace...)

First element to be marked is scheduler

@gvoskuilen can you review before I release to the autotester?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -47,15 +47,15 @@ SST_CHECK_OSX()
 
 SST_ELEMENT_CONFIG_OUTPUT()
 
-# Compile flags come from SST-Core, add or remove extra warnings 
-# depending on the use-picky flag 
+# Compile flags come from SST-Core, add or remove extra warnings
+# depending on the use-picky flag
 SST_CHECK_PICKY
 AS_IF([test "x$use_picky" = "xyes"],
       [WARNFLAGS="-Wall -Wextra"],
       [WARNFLAGS=""])
 CFLAGS="$CFLAGS $WARNFLAGS"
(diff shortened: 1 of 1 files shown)


allevin:
@gvoskuilen if we wanted to get more fancy, we could put in a string in the `.depricated` file to be displayed in the configuration summary that would tell what release of SST that the element would be removed...  Thoughts?

configure.ac
```diff
@@ -149,34 +149,38 @@ for depfilepath in $list; do
...
 echo "Configuration Information (Make will build the following elements):"
 echo ""
 
 for e in $dist_element_libraries
 do
     echo "$active_element_libraries" | grep "$e" >/dev/null
     if test $? -eq 0 ; then
-        msg="YES"
+        if test -e "./src/sst/elements/$e/.depreciated" ; then
+            msg="YES - DEPRECIATED"
```
gvoskuilen: deprecated is misspelled

gvoskuilen:
@allevin I think it's fine as is, if we come across an instance where we decide to leave something deprecated through multiple major releases we could add that.

configure.ac
allevin: You know I thought that the spelling looked funny.

allevin:
@gvoskuilen ok, I fixed the spelling, thanks for the catch!  👍

PR #1428, opened 2020-02-13
Modify makefiles to register source and test paths

allevin:
Register paths of source code and tests directory (for future use) for all elements.  These will be saved into the sstsimulator.conf file when the elements are installed

This change is needed to support the new testing infrastructure.

NOTE: This change should have absolutely no affect on the existing elements operations.


allevin:
@gvoskuilen Can you review this before I release it to the autotester?

PR #1477, opened 2020-04-28
Changes to prospero to support PIN3

allevin:
Disabling compile with zlib if building using PIN3

### DIFF
--- a/src/sst/elements/prospero/proscpu.h
+++ b/src/sst/elements/prospero/proscpu.h
@@ -28,10 +28,8 @@
 #include "prosreader.h"
 #include "prosmemmgr.h"
 
-#ifndef HAVE_PIN3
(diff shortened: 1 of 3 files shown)


src/sst/elements/prospero/proscpu.h
```diff
@@ -28,9 +28,11 @@
 #include "prosreader.h"
 #include "prosmemmgr.h"
 
+#ifndef HAVE_PIN3
```
gvoskuilen, line 31: @allevin This file is not part of the pintool so it can use zlib. It shouldn't have any guards about PIN3.

allevin: Fixed this one

allevin:
Made suggested changes.

PR #1505, opened 2020-06-22
Added subscribe call to other topologies in Merlin

deanchester:
Using the platforms example from @feldergast only Dragonfly topology worked added the subscribe call to the other topologies made this work for other topologies.


feldergast:
These same additions were part of PR #1507.  Closing.

PR #1591, opened 2021-01-18
Update Functional Unit Design in Vanadis and Tidy Up Pipeline Code

nmhamster:
Update provides the following:

- Debug version of Vanadis for when verbose output is required
- Non-debug version of Vanadis for performance runs (`verbose=0` case), compiles out various output generation and self-checks
- Simplifies the functional unit interface and design for slightly improved performance
- Provides a cycle counter for cycles spend pending emulated `SYSCALL` instructions


allevin:
@nmhamster The failures is due to our 10.14 machine going offline.

PR #1602, opened 2021-02-05
Supports unique stdout and stderr per OS instance in Vanadis

nmhamster:
Support in Vanadis emulated OS for unique stdout and stderr files.


allevin:
@nmhamster Make sure your forked branch is up to date with the devel branch.  The err file check on qos dragonfly should not be happening on the latest version from devel.

nmhamster:
> @nmhamster Make sure your forked branch is up to date with the devel branch. The err file check on qos dragonfly should not be happening on the latest version from devel.

@allevin - thanks, I updated it.

PR #1624, opened 2021-03-26
Cleanup  ELI params

mjleven:


allevin:
@mjleven I believe that this PR is intended for devel branch not master?

PR #1662, opened 2021-05-06
fix a bug about writing data to scratchpad

Imsheep:
**Problem:**
After a processor sends a ScratchGet command to the scratchpad and receives a response from the main memory,  the data from the main memory is not written to the scratchpad rightly because of the wrong way of initializing the data vector. 
**Solution:**
The original way of initializing the data vector,
`std::vector<uint8_t> data(response->getPayload()[payloadOffset],response->getPayload()[payloadOffset+size]);` may not copy the response's payload to the data vector correctly.  Because this  form of constructor is matched to the second form, the **fill constructor** about constructing a vector in the [cpp reference about vector](https://www.cplusplus.com/reference/vector/vector/vector/).
So I modified it to match the third one, the **range constructor** in the [cpp reference about vector](https://www.cplusplus.com/reference/vector/vector/vector/).

In this way, it can make my simulation work correct.


gvoskuilen (approved):
Thank you for reporting this bug.

PR #1663, opened 2021-05-06
Fixed issue with rxMatchDelay not getting picked up correctly

deanchester:
This fixes an issue where the rxMatchDelay doesn't get picked up when initialising the NIC in firefly.


feldergast:
I found some other small issues with the merlin python module and have this fix already in with those.  I'll just plan to submit that all as one PR.

feldergast:
Fix included in #1664

PR #1670, opened 2021-05-17
Changing pass/fail test for non-empty error files to be a note

allevin:
Changing pass/fail test for non-empty error files to be a testing note instead with no P/F result


gvoskuilen (approved):
I think this is a good change and will avoid spurious failures but still alert that something unexpected occurred.

PR #1696, opened 2021-07-16
Either a typo or a duplicate key

kistlers:
Either the definition of this value is a typo (rxMemcpyModParams instead of txMemcpyModParams) or the definition of this key is duplicate of the one line 119 above.
I do suspect the first is the case.


nmhamster (approved):

hughes-c:
@nmhamster @mjleven It looks like this defaults to "None" if unset (rangeLatMod.h). If the test is really testing "Mult" then new reference files need to be generated with the correct receive parameter.

hughes-c:
Updated ref files in https://github.com/sstsimulator/sst-elements/pull/1851

Issue #1693, opened 2021-07-14
Vanadis should detect when the binary is MIPS and not MIPSEL


calewis:
Elements sha is 93ba52d

Program: 
```c++ 
#include <cstdio>
int main(){printf("Hello, World!\n");}
```

Compile command:
```
mipsel-linux-musl-g++ -g prog.cpp -static
```

Trying to run Vanadis with this binary copying the basic_vanadis.py file and swapping the executables appears to hang:
```
[Core:   0]: ============================ Cycle        17806 ============================
[Core:   0]: -- Core Status:
[Core:   0]: ---> Thr:     0 (unhalted) / ROB-Pend: 0 / IntReg-Free: 93 / FPReg-Free: 133
[Core:   0]: -- Resetting Zero Registers
[Core:   0]: => Fetch Stage <==========================================================
[Core:   0]: => Decode Stage <==========================================================
[Core:   0]: -> Decode step for thr: 0
[Core:   0]: ---> Max decodes per cycle: 2
[Core:   0]: Instruction Loader - Internal State Report:
[Core:   0]: --> Cache Line Width:          64
[Core:   0]: --> Pending Loads:             1
[Core:   0]: -----> Address:       0xa0014780
[Core:   0]: --> uop Cache Entries:         480 / 1536
[Core:   0]: --> Predecode Cache Entries:   4 / 4
[Core:   0]: ---> uop bundle and pre-decoded bytes are not found (ip=0xa0014780), requesting icache read (line-width=64)
[Core:   0]: [ins-loader] ---> processing a requested load ip=0xa0014780, icache-line: 64 bytes.
[Core:   0]: [ins-loader] ---> issue ins-load line-start: 0xa0014780 (offset=0), line-len: 64 read-len=64 
[Core:   0]: [ins-loader] -----> load is already in progress, will not issue another load.
[Core:   0]: ---> Performed 0 decodes this cycle, 0 uop-bundles used / updated-ip: 0xa0014780.
[Core:   0]: -> Decode step for thr: 0
[Core:   0]: ---> Max decodes per cycle: 2
[Core:   0]: Instruction Loader - Internal State Report:
[Core:   0]: --> Cache Line Width:          64
[Core:   0]: --> Pending Loads:             1
[Core:   0]: -----> Address:       0xa0014780
[Core:   0]: --> uop Cache Entries:         480 / 1536
[Core:   0]: --> Predecode Cache Entries:   4 / 4
[...]

PR #1699, opened 2021-07-19
Vanadis add check for Endian in ELF Format Checks

nmhamster:
Perform a check for the endianness of the executable during ELF load. If this is not a supported type, then fatal. 

This is reported by @calewis in PR #1693.


calewis:
Suggestion, add an enum:

```cpp
enum class MIPSEndian {
Little, 
Big 
};
```
And have the check be `if(elf_info->getEndian() == MIPSEndian::Big){...`, While it's in the error string, it's not clear what `1` means. 

The enum doesn't have be that exactly, but you get the idea.  Alternatively change `getEndian()` to `bool isBigEndian()` then the check is also obvious.

PR #1734, opened 2021-10-27
memH: fix mmap based backing to memory_file

vinayby:
The 'memory_file' parameter description on some components in
memH suggests support for 'storing resulting state' through
to the specified file. This fix mmap's as MAP_SHARED, and ensures munmap/close gets called.


gvoskuilen (approved):
@vinayby Thanks for contributing this.

PR #1737, opened 2021-11-01
memH: Bug fix for incorrect prefetch latency statistics

daecheolyou:
recordPrefetchLatency() doesn't work since there is no startTimes_ map entries for prefetch requests. Unlike handleEvent() in
cacheController, which calls recordIncomingRequest() for an incoming request, handlePrefetchEvent() does not call recordIncomingRequest(), and thus corresponding latency is not collected.

### DIFF
--- a/src/sst/elements/memHierarchy/cacheController.cc
+++ b/src/sst/elements/memHierarchy/cacheController.cc
@@ -62,13 +62,6 @@ void Cache::handleEvent(SST::Event * ev) {
  *  a prefetch target
  */
 void Cache::handlePrefetchEvent(SST::Event * ev) {
-    MemEventBase *event = static_cast<MemEventBase*>(ev);
-
-    // Record the time at which requests arrive for latency statistics
-    if (CommandClassArr[(int)event->getCmd()] == CommandClass::Request &&
-        !CommandWriteback[(int)event->getCmd()])
-        coherenceMgr_->recordIncomingRequest(event);
-
     prefetchSelfLink_->send(prefetchDelay_, ev);
 }
 
@@ -82,6 +75,11 @@ void Cache::processPrefetchEvent(SST::Event * ev) {
     if (!clockIsOn_) {
         turnClockOn();
     }
+
+    // Record the time at which requests arrive for latency statistics
+    if (CommandClassArr[(int)event->getCmd()] == CommandClass::Request &&
+        !CommandWriteback[(int)event->getCmd()])
+        coherenceMgr_->recordIncomingRequest(static_cast<MemEventBase*>(event));
 
     // Record received prefetch
     statPrefetchRequest->addData(1);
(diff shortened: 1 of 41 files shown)


calewis:
@daecheolyou thanks for the PR (PRs are always welcome), I assigned it to @gvoskuilen since she is the main memH dev.

src/sst/elements/memHierarchy/cacheController.cc
```diff
@@ -62,6 +62,13 @@ void Cache::handleEvent(SST::Event * ev) {
  *  a prefetch target
  */
 void Cache::handlePrefetchEvent(SST::Event * ev) {
+    MemEventBase *event = static_cast<MemEventBase*>(ev);
+
+    // Record the time at which requests arrive for latency statistics
+    if (CommandClassArr[(int)event->getCmd()] == CommandClass::Request &&
```
gvoskuilen, line 68: Can you move this to the processPrefetchEvent funciton where the rest of the statistic handling is done? That will start the latency timer at the point at which the prefetch is recorded as 'received'.

daecheolyou: I updated the code as you commented.

gvoskuilen (approved):

daecheolyou:
@gvoskuilen 
It seems that the tests compares the simulation results with reference files (in refFiles). However, the simulation results are bound to be different from the reference files since the chagnes in the pull request reflect the hidden prefetch latency, which was not added to the stat before. Could you pleasse check this?

gvoskuilen:
@daecheolyou This is failing testing because the statistics outputs are affected by the change and so the tests now diff against the reference files. You can either (1) Run sst-test-elements which will tell you which tests are failing and generate new reference files that you'll have to copy into the tests/refFiles directories in the elements or (2) I have a PR I'm putting together with various memH upates and if it's easier, I can make this fix there and fix the reference files. Let me know what you'd prefer.

daecheolyou:
@gvoskuilen If I generate new ref. files, how can I reflect them to SST-autotester?

gvoskuilen:
@daecheolyou The reference files are part of the repository so update the reference files in your repo (mostly in elements/*/tests/refFiles/) and then commit them to the branch. I'll then re-review the PR and submit for testing.

gvoskuilen (approved):

daecheolyou:
@gvoskuilen I added reference files related to Goblin HMC, which were missed in the previous commit (Formerly, I built sst-elements without goblin). However, test_memHA_MemoryCache, PrefetchParams and ThroughputThrottling were updated in the previous commit and related tests also passed in my environment. Could you please check why sst-autotester fails for those tests?

gvoskuilen (approved):

gvoskuilen:
@daecheolyou The failing tests are parallel (MT-2 --> threads=2, MR-2 --> ranks=2) and they are diffing in the latency statistic. These stats shouldn't diff from the serial case however, I'm letting the test run again.

daecheolyou:
@gvoskuilen I found out that when the test is done with multi-threaded option, *_MC.out is used for a ref. file (e.g., test_memHA_PrefetchParams_MC.out instead of test_memHA_PrefetchParams.out). So I updated the related ref. files.

gvoskuilen (approved):

PR #1830, opened 2022-02-24
Update OS Calls for RISC-V Vanadis

nmhamster:
- Adds support for `brk` calls for RISCV
- Initial support for `mmapx` operations that give us more flexible `mmap` modeling


hughes-c:
FAIL -- test_vanadis_short_tests_003_small_basic_io_printf_check

PR #1861, opened 2022-04-05
Fix Vanadis I/O open system call

mjleven:
Vanadis I/O was mapping emulated open/openat system calls to a fopen SST call. As is the code was not handling flags and mode properly and failing for some cases. This PR changed to a direct mapping from emulated open/openat to SST open/openat system calls. This allows flags and mode to be passed directly to SST system calls open/openat. This approach implies the need to convert open flags from architecture of the binary running in Vanadis to values matching the architecture of the SST host. The architecture specific flags where generated by running a utility that prints out the flags.


nmhamster:
@mjleven - looks good, let's get it tested!

PR #1879, opened 2022-04-19
Move code that loads the elf text/data/appEnvironmentData from Vanadis component to OS component

mjleven:
1) Moved code that loads elf text and data from vanadis.cc  to os/velfload.cc
2) Moved code that sets up the applications stack and environment from decoder/vmipsdecoder.h to os/vappruntimememory.h. This code now works for both 32 and 64 bit pointers via a template and is incorporated into a SST module. The correct module is loaded by class VanadisNodeOSComponent() based on elf info.
3) The stackStart and instructionPointer is sent from the OS component to the Vanadis component via an event passed through the OSHandler for thread 0. The path for this is a bit awkward at the moment and will be cleaned up when I single OShandler is used for all decoders.
4) The code in VanadisNodeOSComponent is currently s single point solution and will need to be generalized at some point.


nmhamster:
Let's try the testing for this one.

PR #1882, opened 2022-04-20
Add Additional Debug Capabilities to memH

hughes-c:
Added the ability to dump the backing store and see data payload in cache transactions. Warning -- the backing store dump can be *large*.

### DIFF
--- a/src/sst/elements/memHierarchy/lineTypes.h
+++ b/src/sst/elements/memHierarchy/lineTypes.h
@@ -283,9 +283,6 @@ class L1CacheLine : public CacheLine {
             std::ostringstream str;
             str << "LLSC: " << (LLSCAtomic_ ? "Y" : "N");
             str << " Lock: (" << userLock_ << "," << eventsWaitingForLock_ << ")";
-            str << " Data: ";
-            for( auto i : data_ )
-                str << uint32_t(i) << " ";
             return str.str();
         }
 };
--- a/src/sst/elements/memHierarchy/lineTypes.h
+++ b/src/sst/elements/memHierarchy/lineTypes.h
@@ -283,6 +283,9 @@ class L1CacheLine : public CacheLine {
             std::ostringstream str;
             str << "LLSC: " << (LLSCAtomic_ ? "Y" : "N");
             str << " Lock: (" << userLock_ << "," << eventsWaitingForLock_ << ")";
+            str << " Data: ";
+            for( auto i : data_ )
+                str << uint32_t(i) << " ";
             return str.str();
         }
 };
--- a/src/sst/elements/memHierarchy/memoryController.cc
+++ b/src/sst/elements/memHierarchy/memoryController.cc
@@ -577,7 +577,8 @@ void MemController::finish(void) {
         memBackendConvertor_->turnClockOn(cycle);
     }
 
-    if( dbg.getVerboseLevel() >= 4 ) {
+    // Print backing store contents
+    if( dbg.getVerboseLevel() > 10 ) {
         backing_->printContents(dbg);
     }
 
@@ -594,12 +595,6 @@ void MemController::writeData(MemEvent* event) {
         }
 
         backing_->set(addr, event->getSize(), event->getPayload());
-        if (is_debug_event(event)) {
-            Debug(_L10_, "\tContents ");
-            for( auto i : event->getPayload() )
-                std::cerr << uint32_t(i) << " ";
-            std::cerr << "\n";
-        }
 
         return;
     }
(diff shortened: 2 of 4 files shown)


src/sst/elements/memHierarchy/memoryController.cc
```diff
@@ -589,6 +594,12 @@ void MemController::writeData(MemEvent* event) {
         }
 
         backing_->set(addr, event->getSize(), event->getPayload());
+        if (is_debug_event(event)) {
+            Debug(_L10_, "\tContents ");
+            for( auto i : event->getPayload() )
+                std::cerr << uint32_t(i) << " ";
+            std::cerr << "\n";
+        }
```
gvoskuilen, line 602: The call to printDataValue() above (line 593) prints the payload if debug level is 11. Is that not enough? Same with below (line 611 prints)

src/sst/elements/memHierarchy/memoryController.cc
```diff
@@ -601,6 +612,12 @@ void MemController::writeData(MemEvent* event) {
         }
         
         backing_->set(addr, event->getSize(), event->getPayload());
+        if (is_debug_event(event)) {
+            Debug(_L10_, "\tContents ");
```
gvoskuilen, line 616: Can you change to _L11_? The rest of memH prints values if debug is 11 or higher so it'll be consistent.

src/sst/elements/memHierarchy/memoryController.cc
```diff
@@ -696,6 +713,12 @@ void MemController::processInitEvent( MemEventInit* me ) {
         if (is_debug_event(me)) { Debug(_L9_,"Memory init %s - Received Write for %" PRIx64 " size %zu\n", getName().c_str(), me->getAddr(),me->getPayload().size()); }
         if ( isRequestAddressValid(addr) && backing_ ) {
             backing_->set(addr, me->getPayload().size(), me->getPayload());
+            if (is_debug_event(me)) {
+                Debug(_L10_, "\tContents ");
```
gvoskuilen, line 717: print at `_L11_` instead.

src/sst/elements/memHierarchy/lineTypes.h
```diff
@@ -283,6 +283,9 @@ class L1CacheLine : public CacheLine {
             std::ostringstream str;
             str << "LLSC: " << (LLSCAtomic_ ? "Y" : "N");
             str << " Lock: (" << userLock_ << "," << eventsWaitingForLock_ << ")";
+            str << " Data: ";
+            for( auto i : data_ )
+                str << uint32_t(i) << " ";
```
gvoskuilen, line 288: Is this just so that the printStatus includes data values? I'd almost rather add a second function since this one is used all over debug code which also prints data values if the debug level is high enough (but the values aren't printed via this function). It'll duplicate values in the output.

hughes-c: Yeah, it's useful to be able to see what's happening. I didn't see this functionality anywhere else -- you say there is a way to print data values? I'm fairly certain that I turned the debug verbosity up to 99 when I was testing this...granted this is two years old now.

src/sst/elements/memHierarchy/memoryController.cc
hughes-c: Let me go look. I wrote this ~2y ago and I didn't see this functionality then.

src/sst/elements/memHierarchy/lineTypes.h
gvoskuilen: I added it fairly recently...so it didn't exist two years ago. I don't think I've caught all the components yet but the MESI/MSI caches and the memory controller print data values. 

I do think we should merge the piece that adds printing the full state of the backing store. And the caches don't print values in printStatus nor is it possible to trigger it on finish() right now, which is a useful capability too. Aside from debug, memH has "verbose". If verbose is set to 3 the state of components get dumped on an SST fatal() call. I think we should add data values to that dump (and make it apply to the memory backing store too) and also have an option that if verbose=4, the state prints on finish(), even if debug isn't enabled. What do you think? Would that give you all the output you need?

hughes-c: Removed all modifications to lineTypes

src/sst/elements/memHierarchy/memoryController.cc
hughes-c: Removed

src/sst/elements/memHierarchy/memoryController.cc
hughes-c: Removed

PR #1917, opened 2022-06-08
memH: Bug fix in MESI_L1

TaekyungHeo:
Added missing parentheses in `MESI_L1.cc`


gvoskuilen (approved):
Looks good, thanks for the fix!

PR #1954, opened 2022-10-24
Use proper asm for linux/osx

jpkenny:
Mercury bug fix: Use proper asm for linux/osx

### DIFF
--- a/src/sst/elements/mercury/Makefile.am
+++ b/src/sst/elements/mercury/Makefile.am
@@ -39,6 +39,7 @@ libhg_la_SOURCES = \
 #  threading/threading_pth.cc \
 #  threading/threading_ucontext.cc
 
+if SST_COMPILE_OSX
 if SST_HG_X86
   libhg_la_SOURCES += \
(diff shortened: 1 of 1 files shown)


jleidel:
This is a critical PR as it prevents us from testing the BFS motif on the `devel` branch.

jpkenny:
> This is a critical PR as it prevents us from testing the BFS motif on the `devel` branch.

Apologies. We've been recovering from issues with our testing. This should fix the issue on devel. I'll make one final test on my Ubuntu 18.04 system when I get a chance before closing the issue.

feldergast:
> > This is a critical PR as it prevents us from testing the BFS motif on the `devel` branch.
> 
> Apologies. We've been recovering from issues with our testing. This should fix the issue on devel. I'll make one final test on my Ubuntu 18.04 system when I get a chance before closing the issue.

If you ever run into an issue in the future where an element library you aren't using should happen to not compile, you can just put a .ignore file in the element library directory and rerun autogen.sh and it will exclude that directory from compiling.

PR #1957, opened 2022-10-28
Vanadis: add new instructions and update tests

mjleven:
1) Fix a bug where the speculative load address would wrap around
2) Fix a bug where the ReadResp payload was less the load_width
3) Fix a bug where split load payload size was incorrectly set
4) decode instructions that are present but not decoded: FCVT.W.D, FCVT.WU.D, FCVG.L.D, FCVT.LU.D, FCVT.D.L, FCVG.D.LU, 
5) add VanadisFPFusedMultiplyAddInstruction and decode decode it, FMADD.D
6) add new tests gettime and splitLoad
7) enable testing of stream riscv
8) fix GETTIME system call. riscv does not implement GETTIME64 only GETTIME
9) add a name to the ISA tables that will be printed during debug
10) add test test mt-dgemm
11) add VanadisFPFusedMultiplySubInstruction and decode it, FMSUB.D
12) change debug in Vanadis a bit, add the ability to print the ROB when a instruction is issued or retired
13) decode instruction FEQ.D
14) change how FMADD checks for invalid op, check the values and don't use c++ exceptions


src/sst/elements/vanadis/inst/vfpmadd.h
```diff
@@ -0,0 +1,142 @@
...
+        if ( (sizeof(fp_format) == 8) && (VANADIS_REGISTER_MODE_FP32 == isa_options->getFPRegisterMode()) ) {
+            const fp_format src_1  = combineFromRegisters<fp_format>(regFile, phys_fp_regs_in[0], phys_fp_regs_in[1]);
+            const fp_format src_2  = combineFromRegisters<fp_format>(regFile, phys_fp_regs_in[2], phys_fp_regs_in[3]);
+            const fp_format src_3  = combineFromRegisters<fp_format>(regFile, phys_fp_regs_in[4], phys_fp_regs_in[5]);
+            const fp_format result = std::fma( (fp_format) src_1, (fp_format) src_2, (fp_format) src_3 ); 
+
+            if ( std::fetestexcept(FE_INVALID) ) {
+                fpflags.setInvalidOp();
+                update_fp_flags = true;
+            }
```
nmhamster, line 104: Are you doing this to trap exceptions generated by the host? Because you should clear the exception flags before to make sure an exception didn't occur earlier in execution.

nmhamster: I think it is better to perform the checks against values than look at the exception flags on the host.

nmhamster (changes requested):

nmhamster (approved):

PR #1962, opened 2022-11-05
Vanadis mainly adding new RISCV instructions

mjleven:
- add SLT, MULHU, REM, FSFLAGS, FSQRT.D, FMIN.D, FMV.X.W, FNMSUB.D
- in the LSQ drop all loads that are out of the address range
- add RISCV hpcg binary
- create a new sst runtime config that does not use a router, connect the L2 to directory controller
- add system call support to RISCV 
- update gold files


src/sst/elements/vanadis/decoder/vriscv64decoder.h
```diff
@@ -634,8 +646,15 @@ class VanadisRISCV64Decoder : public VanadisDecoder
...
-                        // MULHSU
-                        // NOT SURE WHAT THIS IS?
+                        // MULHU mul und place in upper XLEN bits, unsigned 
+                        // need to check the register order
+                        output->verbose(
+                            CALL_INFO, 16, 0, "-------> MULHU %" PRIu16 " <- %" PRIu16 " + %" PRIu16 "\n", rd, rs1, rs2);
+
+                        bundle->addInstruction(
+                            new VanadisMultiplyHighInstruction<uint64_t,uint64_t,uint64_t,__uint128_t,false,false>(
+                                ins_address, hw_thr, options, rd, rs1, rs2));
```
nmhamster, line 656: I don't think there are any 64bit operations with `MULH` series of instructions, they are all 32bit (I think)? I think this operation should be `uint32_t` based?

src/sst/elements/vanadis/decoder/vriscv64decoder.h
```diff
@@ -712,6 +730,16 @@ class VanadisRISCV64Decoder : public VanadisDecoder
...
                         decode_fault = false;
                     } break;
+                    case 0x1:
+                    {
+                        // REM
+                        output->verbose(
+                            CALL_INFO, 16, 0, "-----> REM %" PRIu16 " <- %" PRIu16 " %% %" PRIu16 "\n", rd, rs1, rs2);
+                        bundle->addInstruction(
+                            new VanadisModuloInstruction<int64_t>(
+                                ins_address, hw_thr, options, rd, rs1, rs2));
```
nmhamster, line 740: `REM` and `REMU` are 32bit instructions.

src/sst/elements/vanadis/inst/vfpsqrt.h
```diff
@@ -0,0 +1,116 @@
...
+            isa_fp_regs_out[0] = dest;
+        }
+    }
+
+    VanadisFPSquareRootInstruction* clone() override { return new VanadisFPSquareRootInstruction(*this); }
+    VanadisFunctionalUnitType     getInstFuncType() const override { return INST_FP_ARITH; }
+
+    const char* getInstCode() const override
+    {
+        if ( std::is_same<fp_format, double>::value ) { return "FPSQRT"; }
```
nmhamster, line 58: This should be `FP64SQRT`

src/sst/elements/vanadis/inst/vmulhigh.h
```diff
@@ -0,0 +1,88 @@
...
+        isa_int_regs_in[1]  = src_2;
+        isa_int_regs_out[0] = dest;
+    }
+
+    VanadisMultiplyHighInstruction* clone() override { return new VanadisMultiplyHighInstruction(*this); }
+
+    VanadisFunctionalUnitType getInstFuncType() const override { return INST_INT_ARITH; }
+    const char*               getInstCode() const override
+    {
+        if( signed_left && signed_right) {
```
nmhamster, line 44: Can you set these so they are `MULH32` etc? In RISCV I think `MULH`, `MULHU` and `MULHSU` are all 32bit width results (even though the multiplication is performed by a 64-bit multiply to extract the higher 32 bits.

src/sst/elements/vanadis/lsq/vbasiclsq.h
```diff
@@ -84,7 +84,7 @@ class VanadisBasicLoadStoreQueue : public SST::Vanadis::VanadisLoadStoreQueue {
             new StandardMem::Handler<SST::Vanadis::VanadisBasicLoadStoreQueue>(
                 this, &VanadisBasicLoadStoreQueue::processIncomingDataCacheEvent));
 
-        address_mask = params.find<uint64_t>("address_mask", 0xFFFFFFFFFFFFFFFFULL);
+        address_mask = params.find<uint64_t>("address_mask", 0xFFFFFFFFULL);
```
nmhamster, line 87: The default should be full 64bits shouldn't it? This would be the equivalent to an empty mask. Once we have VM working, we don't want any mask right?

nmhamster (changes requested):

src/sst/elements/vanadis/decoder/vriscv64decoder.h
hughes-c: Are you sure? XLEN should be 64. I think REM[U]W are the 32b variations for RV64.
Format
    rem rd,rs1,rs2
Description
    perform an XLEN bits by XLEN bits signed integer reminder of rs1 by rs2.
Implementation
    x[rd] = x[rs1] %s x[rs2]

src/sst/elements/vanadis/inst/vmulhigh.h
hughes-c: The same comment here. XLEN is 64, so it should produce a 128b result and put the upper 64b in the destination register.

PR #1991, opened 2022-11-30
2D BFS Models

plavin:
This PR updates the Ember BFS motifs to use 2D models.


plavin:
This PR contains an SDL that uses this new file format:

https://github.com/sstsimulator/a-sst/pull/1

Issue #1916, opened 2022-06-06
memNIC passes the wrong parameters to linkcontrol


tomhaber:
On lines 
https://github.com/sstsimulator/sst-elements/blob/47f2a11e34c5d45d3c3db7f9ca1f9051dd1c1574/src/sst/elements/memHierarchy/memNIC.cc#L44
and
https://github.com/sstsimulator/sst-elements/blob/47f2a11e34c5d45d3c3db7f9ca1f9051dd1c1574/src/sst/elements/memHierarchy/memNIC.cc#L45

It should be "input_buf_size" and "output_buf_size"

jleidel:
@tomhaber, this has been merged into `devel`.  I would wait until the next merge to `master` before you attempt to utilize the feature.

PR #1998, opened 2022-12-16
fixing merlin linkcontrol parameters for memHierarchy.memNIC

jleidel:
@tomhaber submitted a bug (Issue #1916) regarding the `memHierarchy.memNIC` coupling to `merlin.linkcontrol`.  The bug is associated with the correct naming conventions for the `merlin.linkcontrol` `input_buf_size` and `output_buf_size`.  This is a very simple fix to inject the correct names in the subcomponent parameter setup for this particular subcomponent.  

Version tested: sst-elements-devel (2022.12.16); sst-core-devel (2022.12.16)

### DIFF
--- a/src/sst/elements/memHierarchy/memNIC.cc
+++ b/src/sst/elements/memHierarchy/memNIC.cc
@@ -47,7 +47,7 @@ MemNIC::MemNIC(ComponentId_t id, Params &params, TimeConverter* tc) : MemNICBase
         std::string link_control_class = params.find<std::string>("network_link_control", "merlin.linkcontrol");
 
         if (link_control_class != "merlin.linkcontrol")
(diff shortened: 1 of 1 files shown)


gvoskuilen (commented):
@jleidel "in_buf_size" works for kingsley.linkcontrol and "input_buf_size" works for "merlin.linkcontrol". So this will fix the default case but break the kingsley case. Would you mind also changing line 50 from a warning to a fatal error? Something like `dbg.fatal(CALL_INFO, -1, "%s, Error: use of the 'network_link_control' parameter is no longer supported. Please specify a named 'linkcontrol' subcomponent in the input configuration.\n", getName().c_str());`

gvoskuilen (approved):
Thanks!

PR #2005, opened 2022-12-23
memH: Avoid duplicate entries in retryBuffer_

TaekyungHeo:
This pull request avoids duplicate entries in `retryBuffer_`. The retry buffer in caches is used for queueing events to retry. The retry buffer is designed under the assumption that all events in the buffer are unique. When there are duplicate events in the retry buffer, SST may fail to find an address in MSHR ("Error: MSHR::removePendingRetry. Address does not exist in MSHR."). Duplicate events in the retry buffer may appear when a cache is shared and multiple requests for the same address arrive in the same cycle. This patch tracks the address set in the retry buffer, and when an address is found in the address set, the corresponding event is not added to the retry buffer. This logic is only applied to shared caches because private caches will not experience the case where the same address is accessed by multiple requests at the same time.

### DIFF
--- a/src/sst/elements/memHierarchy/coherencemgr/Incoherent.cc
+++ b/src/sst/elements/memHierarchy/coherencemgr/Incoherent.cc
@@ -478,6 +478,7 @@ bool Incoherent::handleNULLCMD(MemEvent* event, bool inMSHR) {
     if (evicted) {
         notifyListenerOfEvict(line->getAddr(), lineSize_, event->getInstructionPointer());
         retryBuffer_.push_back(mshr_->getFrontEvent(newAddr));
+        retryBufferAddrSet_.insert(newAddr);
         if (mshr_->removeEvictPointer(oldAddr, newAddr))
             retry(oldAddr);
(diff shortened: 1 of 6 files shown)


gvoskuilen:
@TaekyungHeo I'm looking into what could allow multiple requests to the same address in the same cycle. Generally the arbitrateAccess() function in the cacheController should prevent multiple requests for the same address from being handled in the same cycle, which should prevent this from happening unless a two requests to different addresses are triggering retries of other requests for the same address which is possibly a deeper issue in the coherence logic. Can you describe the sequence of events that leads to the bug? Thanks!

TaekyungHeo:
Thanks for reviewing my codes. I experienced this issue while working on a Georgia Tech internal project. We assume a system with multi-core and prefetchers. I implemented a prefetcher in caches. Maybe the prefetchers that I have implemented are causing the problem. I will reopen this issue when I can confirm that this is a common problem for all SST users.

gvoskuilen:
@TaekyungHeo Thanks! I will also scan through the prefetch-handling code in case there is an issue there.

Issue #1799, opened 2022-01-17
Documentation for SST-elements Ember, Firefly and Merlin


saichenna:
We are currently using SST to evaluate the performance of some existing (and a couple of notional) exascale network topologies. As part of building new topologies and adding new routing algorithms, we are having some difficulties due to lack of sufficient documentation. Some of the resources we found online are slides that we presume are part of some tutorial, thereby making it harder to get enough information. It would be great if you can provide any detailed documentation on the Ember, Firefly and Merlin elements, specifying the general workflow of each component and explanation of some of the key APIs/functions in each component. That would help us understand the network simulator in better detail.

In specific, we would like to know more details about the following:
1. Merlin: 
    a. What are the general steps needed to implement a new topology? Based on our understanding, we need to create a python class (need to have a build method) for this topology and declare it as a subcomponent. We also need to implement the routing logic (in .cc file). We would like to know what are the basic functions needed in the routing class (like route_packer, route_init etc..) and if they are called from other components outside merlin (like Firefly or simulator engine itself). 
    b. A basic FSM/workflow-diagram of how Ember-Firefly-Merlin components interact during the simulation would be really useful.
    c. Also, what is the purpose of a virtual network (VN)?
2. Firefly:
   a. Basic documentation on what Firefly does to the event queue received from Ember before sending it to the Merlin router. What are some of the key functions and their purpose? 
3. Ember:
   a. Basic documentation on how the event queues are generated from the available benchmarks. 
   b. Can you specify more details on emberotf2.cc. motif ?

feldergast:
Here are some answers to your merlin questions:

Implementing a new topology requires two parts.  A C++ SubComponent derived from the Topology class defined in sst/lelements/merlin/router.h.  A python class derived from the Topology class in sst/elements/merlin/pymerlin-base.py.  The purpose of the C++ class is the provide all the functionality to assign routes to packets in the router.  The python class is responsible for creating the connection graph for the simulation during simulation creation.

The C++ topology class is called only by hr_router to help it know how to move packets through the network.  This is a common feature of all SST SubComponents, they are only called by the object that loads them.  I'll try to add some additional comments to the Topology class in router.h to better explain when the functions are called and what they are intended to do.  For now, looking at the hyperx topology class in sst/elements/merlin/topology will give you a good example of what each of the functions does.

[...]

PR #2061, opened 2023-03-05
documentation example local to the source code

researcherben:
Example of documentation local to the directory.

Content is from https://github.com/sstsimulator/sst-elements/issues/1799


berquist:
@feldergast @gvoskuilen We discussed in a meeting ~2 weeks ago a preference to not have local documentation and instead stick to
- public code documentation in Doxygen and
- "fully formalized" code and design documentation in http://sst-simulator.org/sst-docs/.

What do you think?  Both the website and this "local" documentation have a roughly equivalent risk of being out-of-date relative to the code, and putting this in Doxygen is better in that regard.  Injecting Doxygen into sst-docs, which would be more readable, might not be easy.

researcherben:
@berquist - @feldergast explained to me in June that README files in the git repo aren't visible to users of SST who don't read source.

Therefore @feldergast is enabling an extended `--help`

I asked @feldergast whether we could put a README in the respective source directories that say "look at the respective extended --help"?
@feldergast said yes to that.

PR #2078, opened 2023-03-21
Update Balar with mmio interface

William-An:
Update Balar as a mmio device, supporting both CUDA API trace-driven execution (with `balarTestCPU`) and Vanadis (break with the TLB unit right now).


src/sst/elements/balar/balar.cc
hughes-c, line 1: Instead of commenting-out unused features, please remove them.

src/sst/elements/memHierarchy/tests/testStdMem-mmio.py
hughes-c, line 1: Please remove this file from the commit. Just checkout the current version from devel and move this test into the balar test directory if it is sufficiently different from the other two tests.

src/sst/elements/vanadis/tests/small/misc/splitLoad/mipsel/splitLoad
hughes-c, line 1: This is a binary file. Why was it included in the commit?

hughes-c (changes requested):

src/sst/elements/vanadis/tests/small/misc/splitLoad/mipsel/splitLoad
William-An: I think it gets included in the commit history accidentally. Should I get it removed?

src/sst/elements/memHierarchy/tests/testStdMem-mmio.py
William-An: Done

src/sst/elements/balar/balar.cc
William-An: Done

William-An: Actually those commented out section are part of the original balar. Do we want to remove all of the old code once and for all?

William-An:
The `balar_event.h` file still stays in the repo as `ariel.cc` depends on it. I am just including it so that no modifications will be done on ariel. All other old balar stuff should be removed now.

hughes-c (approved):

gvoskuilen:
@William-An I suspect the failures will be fixed by (1) merging the latest devel branch into yours or rebasing on the latest devel branch and (2) check that any new files you have added in this PR are also added to a Makefile.am. You can also run "make dist" on your branch and see which files are missing from the distribution.

William-An:
The `make dist` failed as I have changed the name of one configuration file. Now it should be fixed.

hughes-c (approved):

PR #2132, opened 2023-05-17
Polarity topos

kartiklakhotia:
PolarFly and PolarStar are low-diameter network topologies with state-of-the-art scalability. We have added the following for these topologies in Merlin:

1. Generators
2. Routing functions
3. Test files with golden (reference) output

References:

1. Lakhotia, Kartik, Maciej Besta, Laura Monroe, Kelly Isham, Patrick Iff, Torsten Hoefler, and Fabrizio Petrini. "PolarFly: a cost effective and flexible low-diameter topology." In Proceedings of the International Conference on High Performance Computing, Networking, Storage and Analysis, 2022.

4. Lakhotia, Kartik, Laura Monroe, Kelly Isham, Maciej Besta, Nils Blach, Torsten Hoefler, and Fabrizio Petrini. "PolarStar: Expanding the Scalability Horizon of Diameter-3 Networks." arXiv preprint arXiv:2302.07217 (2023).


gvoskuilen:
@kartiklakhotia I ran this through our tester (same as the tests that will run to merge the PR), and it failed. Looks like the issue is that merlin fails because numpy isn't found, even when running existing topologies. Can you fix it so that the existing merlin-based tests still pass without numpy? Thanks!

Here is the error:
```
FATAL: SSTElementPythonModule: Error running PyImport_ExecCodeModule on topology/pymerlin-topo-polarfly.py.  Details follow:
File "topology/pymerlin-topo-polarfly.py", line 29, <module>
ModuleNotFoundError: No module named 'numpy'
```

kartiklakhotia:
> @kartiklakhotia I ran this through our tester (same as the tests that will run to merge the PR), and it failed. Looks like the issue is that merlin fails because numpy isn't found, even when running existing topologies. Can you fix it so that the existing merlin-based tests still pass without numpy? Thanks!
> 
> Here is the error:
> 
> ```
> FATAL: SSTElementPythonModule: Error running PyImport_ExecCodeModule on topology/pymerlin-topo-polarfly.py.  Details follow:
> File "topology/pymerlin-topo-polarfly.py", line 29, <module>
> ModuleNotFoundError: No module named 'numpy'
> ```

I have removed the numpy requirement. Should resolve this error.

kartiklakhotia:
Isn't SST_ELEMENTS_ROOT supposed to be specified for installation?
http://sst-simulator.org/SSTPages/SSTBuildAndInstall_12dot0dot1_SeriesDetailedBuildInstructions/

The reason why this env variable is used - the finite field generator (GF.py)  is imported by both polarfly and polarstar, and the polarfly topology itself is imported by polarstar. These files are all in merlin/topology/ directory and their absolute location while importing in the code, is determined using SST_ELEMENTS_ROOT variable.  

Can you suggest any alternative?

feldergast:
At least some of our users have a system administrator install SST for all users, in which case, the source files will not actually be available.  So, looking for GF.py in the source files is not really feasible.  There are two options.  First,  you can install GF.py with the element library, but you will still need to then figure out how to get the absolute path to it.  The other option is to embed it in the element library python module.  Neither approach is perfect, but the second approach will remove the need for finding the library at runtime, which could be quite challenging to do in a portable way.

kartiklakhotia:
Updated. Finite field generator code is no more imported from the source.

gvoskuilen:
@kartiklakhotia All the tests are passing now except the two new polarstar & polarfly ones. I think we need two last things: (1) we will get SymPy into our test infrastructure so the tests run and (2) can you change these two tests so they skip the test if sympy isn't there? For example, do a try/catch on the sympy import.

Here is the error:
```
Traceback (most recent call last):
  File "/.../sst-elements/src/sst/elements/merlin/tests/polarfly_455_test.py", line 64, in <module>
    system.build() 
  File "pymerlin-base.py", line 803, in build
  File "topology/pymerlin-topo-polarfly.py", line 444, in build
  File "topology/pymerlin-topo-polarfly.py", line 396, in generate
  File "topology/pymerlin-topo-polarfly.py", line 324, in make
  File "topology/pymerlin-topo-polarfly.py", line 44, in __init__
Exception: --> Did not load Sympy Polys Galoistools : module required for field construction
FATAL: SSTPythonModel: Execution of model construction function failed
```

kartiklakhotia:
> @kartiklakhotia All the tests are passing now except the two new polarstar & polarfly ones. I think we need two last things: (1) we will get SymPy into our test infrastructure so the tests run and (2) can you change these two tests so they skip the test if sympy isn't there? For example, do a try/catch on the sympy import.
> 
> Here is the error:
> 
> ```
> Traceback (most recent call last):
>   File "/.../sst-elements/src/sst/elements/merlin/tests/polarfly_455_test.py", line 64, in <module>
>     system.build() 
>   File "pymerlin-base.py", line 803, in build
>   File "topology/pymerlin-topo-polarfly.py", line 444, in build
>   File "topology/pymerlin-topo-polarfly.py", line 396, in generate
>   File "topology/pymerlin-topo-polarfly.py", line 324, in make
>   File "topology/pymerlin-topo-polarfly.py", line 44, in __init__
> Exception: --> Did not load Sympy Polys Galoistools : module required for field construction
> FATAL: SSTPythonModel: Execution of model construction function failed
> ```

I have updated the code to skip polarfly and polarstar tests when ‘sympy’ is not installed, 
so that the test does not output anything and does not encounter the exception . 
Upon running individual tests using sst-test-elements, polarfly and polarstar tests fail in absence of sympy 
because the (blank) output does not match the reference output. I think this should be the desired behavior 
that these tests do not pass when the topologies cannot be built.

Please let me know if this is fine.

gvoskuilen:
@kartiklakhotia This didn't pass testing due to syntax errors. The issue looks to be the ":" at the end of the `@skipIf` lines.

PR #2158, opened 2023-06-21
Vanadis: Fix syscall to dead process

mjleven:
When an exitgroup happens there is a race condition where a syscall from an hardware thread could be processed by the OS after it terminated the process.


jpkenny:
Moved to WIP until the ARM builder gets fixed.

PR #2211, opened 2023-08-25
Fix MemH MemRtrEvent memory leak

plavin:
This PR fixes a memory leak in MemH. When MemRtrEvent wraps an event and gets deleted, it does not delete the wrapped event. This PR adds a destructor to ensure the wrapped event gets deleted and changes the wrapped object to be protected instead of public. It also includes functions for interacting with the event.
- `takeEvent()` returns the pointer to the event and sets the pointer to `nullptr`. 
- `inspectEvent()` returns the pointer but does not set the pointer to `nullptr`
- `putEvent()` is used to set the value of the Event


gvoskuilen (approved):
Looks good.

Issue #2224, opened 2023-09-25
Pipeline stages sequence in Vanadis


Connie120:
Hello,

I'm wondering why the simulation sequence for the pipeline stages in Vanadis is Fetch->Decode->Issue->Execute->Retire, rather than Retire->Execute->Issue->Decode->Fetch, which is the other way around. 

I tried to run a simple program using the latest Vanadis in the devel branch and printed out the ROB states:

![Screenshot from 2023-09-08 15-15-27](https://github.com/sstsimulator/sst-elements/assets/54909364/6b71507e-b635-4714-91d2-1a0bdc3a9029)

This is the first cycle (cycle 13590) that Vanadis starts to simulate the program. As from the screenshot, the instruction PCADDI64 (Addr 1013e) is decoded and issued in the same cycle, and it's sent to the execution unit. The reason why it's not retired is that it hasn't finished execution. 

So I tried to invert the sequence of the stages and printed out the the ROB states again, also cycle 13590:

![Screenshot from 2023-09-25 15-10-15](https://github.com/sstsimulator/sst-elements/assets/54909364/a3f9d20e-f76a-4088-9299-50c5ffcbdec8)

Here the instruction is only decoded but not issued, and it's issued in the next cycle (cycle 13591):

![Screenshot from 2023-09-25 14-57-38](https://github.com/sstsimulator/sst-elements/assets/54909364/ffa9dd2e-b049-464e-be83-82440e47fd82)

Please let me know if I understand the code incorrectly and thanks for the help!

hughes-c:
@Connie120 After internal discussions, we think that swapping the pipeline order is a viable solution. You're welcome to submit a PR to fix this or we can work on it.

Connie120:
Created a PR in #2240, please check. Thanks!

PR #2240, opened 2023-10-11
Inverted the pipeline stages to Retire->Execute->Issue->Decode->Fetch

Connie120:
This PR invert the pipeline stages from Fetch->Decode->Issue->Execute->Retire to Retire->Execute->Issue->Decode->Fetch. The related issue is in #2224. The changes only happen in the tick() function, and it passes some simple apps. Please check whether more complicated apps would work with this fix. Thanks!

### DIFF
--- a/src/sst/elements/vanadis/vanadis.cc
+++ b/src/sst/elements/vanadis/vanadis.cc
@@ -1235,35 +1235,56 @@ VANADIS_COMPONENT::tick(SST::Cycle_t cycle)
         }
     }
 
-    // Fetch
+    #ifdef VANADIS_BUILD_DEBUG
+    if(output_verbosity >= 9) {
+        output->verbose(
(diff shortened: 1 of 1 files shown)


hughes-c (approved):

gvoskuilen:
These changes cause the Vanadis tests to diff. Re-generate the output files (run `sst-test-elements -w "*vanadis*"`), and copy them from the sst-test-output directory to the appropriate directory in vanadis/tests/. You will also need to update the rdmaNic test files (`sst-test-elements -w "*rdmaNic*"`)

hughes-c:
@Connie120 If you don't want to regenerate the tests, we can cancel this PR and do it internally.

Connie120:
@hughes-c Sure! I'll close this PR.

PR #2255, opened 2023-10-22
fix format specifiers

jmlapre:
Fix format specifiers for all elements except vanadis.  Eric has been working on vanadis warnings.


jmlapre:
Actually it looks like Eric took care of all of them!

PR #2300, opened 2024-01-12
Add to stats more efficiently in various turnClockOn() functions

stkaplan:
I've actually seen these stats take a good chunk of SST's runtime in our simulations, so this should help.


gvoskuilen (approved):
Thanks!

Issue #2319, opened 2024-02-14
basic_vanadis.py simulation hanging


plafratt:
New Issue for sst-elements
--------------------------

1 - Detailed description of problem or enhancement

A simulation run with `src/sst/elements/vanadis/tests/basic_vanadis.py` hangs.

2 - Describe how to reproduce

Edit the configuration file `src/sst/elements/vanadis/tests/basic_vanadis.py` to set `numCpus` to 2 and `numThreads` to 2.

Using attached executable, complete the following steps. [xhpcg.zip](https://github.com/sstsimulator/sst-elements/files/14288506/xhpcg.zip)

```
$> export VANADIS_EXE=./xhpcg
$> export VANADIS_EXE_ARGS"=--rt=1 --nx=1 --ny=1 --nz=1"
$> sst ./basic_vanadis.py
```

After a few minutes, examine the file `hpcg.*.txt`. The program's output in this file indicates that the program is no longer progressing.

3 - What Operating system(s) and versions 

Rocky Linux 8.9

4 - What version of external libraries (Boost, MPI)

5 - Provide sha1 of all relevant sst repositories (sst-core, sst-elements, etc)

sst-elements `7e67f8f29`
sst-core `e952a81bcb`

6 - Fill out Labels, Milestones, and Assignee fields as best possible

plafratt:
In case it is helpful - I looked into the logs, and it appears that all 4 threads are stuck in a futex system call.

Core 0, Thread 0
```
SimStatus: 0:73170703155:---> rob[    0]: addr: 0x000b3ee2 /    SYSCALL / spec:  no / err:  no / issued: yes / front: yes / exe:  no
```

Core 0, Thread 1
```
SimStatus: 0:73170703155:---> rob[    0]: addr: 0x000b3ee2 /    SYSCALL / spec:  no / err:  no / issued: yes / front: yes / exe:  no
```

Core 1, Thread 0
```
SimStatus: 0:73170703155:---> rob[    0]: addr: 0x000b3ee2 /    SYSCALL / spec:  no / err:  no / issued: yes / front: yes / exe:  no
```

Core 1, Thread 1
```
SimStatus: 0:73170703155:---> rob[    0]: addr: 0x000b3ee2 /    SYSCALL / spec:  no / err:  no / issued: yes / front: yes / exe:  no
```

I turned on logging for the OS to identify the system calls.

The last log entries for core 0 (at the end of the log).
```
[...]

PR #2341, opened 2024-03-20
Update futex wake to wake multiple waiters.

plafratt:
Update futex wake to wake all waiters instead of just one. Issue #2319.

### DIFF
--- a/src/sst/elements/vanadis/os/syscall/futex.cc
+++ b/src/sst/elements/vanadis/os/syscall/futex.cc
@@ -16,7 +16,9 @@
 #include <sst_config.h>
 
 #include "os/syscall/futex.h"
+#include "os/callev/voscallfutex.h"
 #include "os/vnodeos.h"
+#include "os/voscallev.h"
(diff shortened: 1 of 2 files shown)


mjleven (changes requested):
I believe you are waking all waiters. The futex man page says to wake "val" number of waiters. This is typically 1 or INT_MAX but could be some value in between.

plafratt:
> I believe you are waking all waiters. The futex man page says to wake "val" number of waiters. This is typically 1 or INT_MAX but could be some value in between.

Thank you for pointing this out. I'd forgotten to cover that part of the specification.

I have a meeting tonight to prepare for, but I will try to address this tomorrow.

Thanks Again,
Patrick

berquist:
Apologies for the delay, we were trying to understand some new CI strangeness.

plafratt:
> Apologies for the delay, we were trying to understand some new CI strangeness.

No worries - thank you for the updates!

PR #2372, opened 2024-06-29
[vanadis] Implement the missing Zicntr extension instructions.

ldalessa:
Chapter 8 in
https://github.com/riscv/riscv-isa-manual/releases/download/riscv-isa-release-f797123-2024-06-27/riscv-unprivileged.pdf details the RISC-V counter interface.

There are 3 bespoke counters detailed in the Zicntr extensions in 8.1, and up to 29 more user-programmable counters detailed in the Zihpm extenstion in 8.2. This commit provides space for all 32 counters in the register file, but only currently implements the three Zicntr counters.

The implementation consists of a few changes.

1. I have extended the register file structure with 32, 64 bit counters, and `increment` and `get` members in order to update and read those counters.
2. I have added a new instruction, `Zicntr::VanadisReadCounterInstruction` in order to read those counters. This instruction can be tailored for any of the 3 `Zicntr`s (implemented as tag types in `inst/zicntr.h`), XLEN=64 or 32, and with or without the `[H]` extension.
3. I have added decoding to `decoder/vriscv64decoder.h` to handle `RDCYCLE`, `RDTIME`, and `RDINSTRET`. Because the `[H]` and XLEN=32 versions are only available in `riscv32`, the `riscv64` decoder simply injects a decoding failure if those occur.
4. I have extended the core's `tick` to update the three `Zicntr`s.
5. I have added a small misc test example to demonstrate usage.

I have not implemented any functionality for programming the 29 `Zihpm` counters, nor have I added any special decoding to read these registers.

I have decided that the `Zicntr` instructions will be processed by the arithmetic functional unit. I don't know if this is appropriate, but it should manage any register port contention properly and seemed like the most expedient solution.

I have removed the `cycle_count` and the `getCycleCount` from the decoder base class, as it is no longer being used. The current cycle is still passed to the decoder `tick()` even though it is _also_ still unused.

Erase this information and put your Pull Request comments here 
---

1 - Verify that the Pull Request is targeted to the **devel** branch of sstsimulator/sst-elements

2 - Verify that Source branch is up to date with the devel branch of sst-elements

3 - After submitting your Pull Request:
   * Automatic Testing will commence in a short while 
      * Pull Requests will be tested with the devel branches of the sst-core and sst-sqe repositories
         * These branches are syncronized with the devel branch of sst-elements.  This is why is it important to keep your source branch up to date.
      * If testing passes, the source branch will be automatically merged (if possible)
         * Pull Requests from forks will not be automatically tested until the code is inspected.
         * Pull Requests from forks will not be automatically merged into the devel branch.
      * If testing fails, You will be notified of the test results.  
         * The Pull Request will be retested on a regular basis - Changes to the source branch can be made to correct problems
         
4 - DO NOT DELETE THE BRANCH (OR FORKED REPO) UNTIL THE PULL REQUEST IS MERGED.
----


ldalessa:
Is there someone that can approve this workflow?

gvoskuilen:
@ldalessa We'll get this reviewed. Our PR testing infrastructure is in the midst of upgrades and so test/merge will be delayed until it is back up.

src/sst/elements/vanadis/decoder/vriscv64decoder.h
```diff
@@ -1107,16 +1105,29 @@ class VanadisRISCV64Decoder : public VanadisDecoder
...
+                            
+                            case 0xc02: // RDINSTRET
+                                bundle->addInstruction( new VanadisReadCounterInstruction( INSTRET, ins_address, hw_thr, options, rd ) );
+                                decode_fault = 0 != rs1;
+                                break;
+
+                            case 0xc80: // RDCYCLEH                            
+                            case 0xc81: // RDTIMEH                            
+                            case 0xc82: // RDINSTRETH
+                                output->verbose( CALL_INFO, 16, 0, "riscv64 does not support Zicntr [H] suffix" );
```
hughes-c, line 1129: Should this kill the simulation?

src/sst/elements/vanadis/decoder/vriscv64decoder.h
```diff
@@ -128,15 +128,13 @@ class VanadisRISCV64Decoder : public VanadisDecoder
     }
 
 
-    void tick(SST::Output* output, uint64_t cycle) override
+    void tick(SST::Output* output, uint64_t) override
```
hughes-c, line 131: Are we better off changing the tick call rather than leaving the uint64_t?

src/sst/elements/vanadis/decoder/vdecoder.h
```diff
@@ -215,7 +215,6 @@ class VanadisDecoder : public SST::SubComponent
     void setThreadLocalStoragePointer(uint64_t new_tls) { tls_ptr = new_tls; }
 
     uint64_t getThreadLocalStoragePointer() const { return tls_ptr; }
-    uint64_t getCycleCount() const { return cycle_count; }
```
hughes-c, line 218: Removing this would seem to break the MIPS32 implementation (see the change to the MIPS decoder). Is there a MIPS equivalent to Zicntr?

src/sst/elements/vanadis/inst/regfile.h
```diff
@@ -52,6 +52,7 @@ class VanadisRegisterFile
     void init( ) {
         std::memset(int_reg_storage, 0, (int_reg_width * count_int_regs));
         std::memset(fp_reg_storage, 0, (fp_reg_width * count_fp_regs));
+        std::fill_n(counters, sizeof(counters), 0);
```
hughes-c, line 55: @feldergast  Is fill_n supported by our current set of compilers (ranges::fill_n looks like c++-20)?

src/sst/elements/vanadis/inst/vzicntr.h
hughes-c, line 1: The RDCYCLE pseudoinstruction reads the low XLEN bits of the cycle CSR which holds a count of
the number of clock cycles executed by the processor core on which the hart is running from an
arbitrary start time in the past. 

The RDTIME pseudoinstruction reads the low XLEN bits of the "time" CSR, which counts wall-clock
real time that has passed from an arbitrary start time in the past. 

The RDINSTRET pseudoinstruction reads the low XLEN bits of the instret CSR, which counts the
number of instructions retired by this hart from some arbitrary start point in the past. 

I'm concerned about the arbitrary bits from the definition in the standard. The example in the manual doesn't show a way to initialize the registers, so I need to do some more digging.

src/sst/elements/vanadis/inst/vzicntr.h
```diff
@@ -0,0 +1,17 @@
...
+#ifndef _H_VANADIS_ZICNTR
+#define _H_VANADIS_ZICNTR
+
+#include <utility>
+
+namespace SST::Vanadis
+{
+    // Tags to be used to with the regfile and VanadisReadCounterInstruction
+    namespace Zicntr {
+        inline constexpr std::integral_constant<uint32_t, 0> CYCLE;
```
hughes-c, line 11: @feldergast Is integral_constant currently supported by our compilers?

hughes-c (commented):
I need to construct another test case for RV64 and do some more digging on the 'arbitrary' piece of the standard. 

The biggest holdup is on the MIPS32 side since you've removed the cycle count. We need to see if there's a way to keep this.

src/sst/elements/vanadis/decoder/vriscv64decoder.h
ldalessa: I assumed that it would be appropriate for this to be considered a `decode_fault` with the verbose output giving a little more information up-front. I'm not particularly tied to any handling here. I'm not sure what the specification has to say about encountering instructions that aren't implemented/able given the implementation's ABI.

src/sst/elements/vanadis/decoder/vriscv64decoder.h
ldalessa: I wasn't sure how invasive to be here. There's no reason for the cycle counter in the tick here for RISC-V in the patch... `cycle` is implemented in the register file, but I wasn't sure what you wanted to do for MIPS (it's not used their either). Removing it wouldn't impact the current code.

src/sst/elements/vanadis/decoder/vdecoder.h
ldalessa: Odd, I didn't see any use if this in the MIPS decoder. I wonder how I missed it. This change can be reverted, it was more of a bookkeeping "dead-code removal" choice and if that's wrong then obviously it can stick around. Or as you say I guess the MIPS register file could also have a cycle counter.

src/sst/elements/vanadis/inst/regfile.h
ldalessa: The ranges version would be 20, but this is the iterator/size version (where iterator here is the decayed pointer to the C-array), and I think that's been around forever.

src/sst/elements/vanadis/inst/vzicntr.h
ldalessa: My understanding is that these are always relative values, and that the absolute value is unimportant. I think you can initialize them to whatever you'd like, as long as they won't overflow.

hughes-c: I think that you're right. I don't see any instruction in the Zi group to explicitly reset these (like resetcycle or something). So this should be fine.

src/sst/elements/vanadis/decoder/vdecoder.h
hughes-c: I don't think that it's used anywhere right now but providing a way to get the cycle count for RV64 and not MIPS32 breaks the template.

This may take some internal discussion to resolve.

ldalessa:
> I need to construct another test case for RV64 and do some more digging on the 'arbitrary' piece of the standard.
> 
> The biggest holdup is on the MIPS32 side since you've removed the cycle count. We need to see if there's a way to keep this.

Thanks for looking at this. I'm pretty sure that the mips decoder did nothing with the cycle count. `getCycleCount` wasn't used there, and all it did was assign `cycle_count = cycle` but `cycle_count` was unused. This was what I saw.

```
ldalessa@sst:~/sandia/sst-elements/src/sst/elements/vanadis$ grep -r getCycleCount
decoder/vdecoder.h:    uint64_t getCycleCount() const { return cycle_count; }
decoder/vriscv64decoder.h:                                    auto thread_call = std::bind(&VanadisRISCV64Decoder::getCycleCount, this);
ldalessa@sst:~/sandia/sst-elements/src/sst/elements/vanadis$ grep -r cycle_count
decoder/vdecoder.h:    uint64_t getCycleCount() const { return cycle_count; }
decoder/vdecoder.h:    uint64_t cycle_count;
decoder/vmipsdecoder.h:        cycle_count = cycle;
decoder/vriscv64decoder.h:        cycle_count = cycle;
```

If you want to retain it in the vmipsdecoder I think it makes sense to make `cycle_count` a member of that class rather than the base class.

src/sst/elements/vanadis/decoder/vdecoder.h
ldalessa: I think if you want to retain this value, even though it's unused currently, a simple solution is to add the `cycle_count` member back into `VanadisMIPSDecoder` rather than the base class, and reinstall the `cycle_count = count` that I removed.

src/sst/elements/vanadis/decoder/vriscv64decoder.h
gvoskuilen: I think this is a reasonable choice; I am also not sure what the specification has to say. The Vanadis config controls whether a `decode_fault` results in a kill.

src/sst/elements/vanadis/inst/vzicntr.h
feldergast: Should be fine.  The last update to integral_constant was in c++14 as far as I can tell.

src/sst/elements/vanadis/inst/vzicntr.h
```diff
@@ -0,0 +1,17 @@
...
+        inline constexpr std::integral_constant<uint32_t, 1> TIME;
+        inline constexpr std::integral_constant<uint32_t, 2> INSTRET;
+    }
```
feldergast, line 14: This really just looks like a class enum type implemented with integral_constant., though I may be missing something deeper.  Is there a reason not to simply use a class enum?

ldalessa: They encode the index as a compile time value wherever it's used. So I use them as a tag type to the `Zicntr::VanadisReadCounterInstruction` class template constructor and I don't have to set any explicit class template parameters. 

```c++
template <uint32_t ID> 
struct Foo {
    Foo(const std::integral_constant<uint32_t, ID>, stuff...);
};

std::integral_constant<uin32_t, 7> TAG;

Foo foo(TAG, stuff...); // ID is inferred as 7
```

as opposed to

```c++
enum class E {
    A, B, ...
};

template <E ID>
struct Foo {
    Foo(stuff...);
};

Foo<E::A> foo(stuff...); // have to be explicit here
```

They're used in the code like:

```c++
bundle->addInstruction( new VanadisReadCounterInstruction( INSTRET, ins_address, hw_thr, options, rd ) )
```

If `vanadis` would prefer an enum and to use `VanadisReadCounterInstruction<Enum::INSTRET>(...)` that's fine with me.

ldalessa: (they're also serve fine as case statement labels)

src/sst/elements/vanadis/inst/vzicntr_readcounter.h
```diff
@@ -0,0 +1,91 @@
...
+            }
+
+            VanadisReadCounterInstruction* clone() override
+            {
+                return new VanadisReadCounterInstruction(*this);
+            }
+    
+            VanadisFunctionalUnitType getInstFuncType() const override
+            {
+                return INST_INT_ARITH;                  // Is this appropriate?
```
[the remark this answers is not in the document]
gvoskuilen: This is a reasonable choice. It'll account for the register use and I doubt consuming an ALU slot will affect the overall reported performance much.

src/sst/elements/vanadis/decoder/vriscv64decoder.h
hughes-c: We should just change the tick() function. However, I'm ok leaving this and fixing it after.

hughes-c:
@ldalessa

hughes-c:
@ldalessa 
`In file included from decoder/vriscv64decoder.cc:19:
./decoder/vriscv64decoder.h: In member function 'void SST::Vanadis::VanadisRISCV64Decoder::decode(SST::Output*, uint64_t, uint32_t, SST::Vanadis::VanadisInstructionBundle*)':
./decoder/vriscv64decoder.h:1112:132: error: class template argument deduction failed:
                                 bundle->addInstruction( new VanadisReadCounterInstruction( CYCLE, ins_address, hw_thr, options, rd ) );
                                                                                                                                    ^
./decoder/vriscv64decoder.h:1112:132: error: no matching function for call to 'VanadisReadCounterInstruction()'
In file included from ./inst/vinstall.h:125,
                 from ./decoder/vriscv64decoder.h:20,
                 from decoder/vriscv64decoder.cc:19:
./inst/vzicntr_readcounter.h:20:13: note: candidate: 'template<unsigned int id, long unsigned int XLEN, bool H> VanadisReadCounterInstruction(std::integral_constant<unsigned int, id>, uint64_t, uint32_t, const SST::Vanadis::VanadisDecoderOptions*, uint16_t)-> SST::Vanadis::Zicntr::VanadisReadCounterInstruction<id, XLEN, H>'
             VanadisReadCounterInstruction(
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./inst/vzicntr_readcounter.h:20:13: note:   template argument deduction/substitution failed:
In file included from decoder/vriscv64decoder.cc:19:
./decoder/vriscv64decoder.h:1112:132: note:   candidate expects 5 arguments, 0 provided
                                 bundle->addInstruction( new VanadisReadCounterInstruction( CYCLE, ins_address, hw_thr, options, rd ) );
                                                                                                                                    ^
./decoder/vriscv64decoder.h:1117:131: error: class template argument deduction failed:
                                 bundle->addInstruction( new VanadisReadCounterInstruction( TIME, ins_address, hw_thr, options, rd ) );
                                                                                                                                   ^
./decoder/vriscv64decoder.h:1117:131: error: no matching function for call to 'VanadisReadCounterInstruction()'`

ldalessa:
Yeah. I’ll take a look at it. It’s likely to be a simple-but-annoying fix.  I don’t realize there was support for such old toolchains.

hughes-c:
Yeah...
http://sst-simulator.org/SSTPages/SSTElementReleaseMatrix/

PR #2373, opened 2024-07-02
Parameter to disable memory region check in MemNICBase

kpgriesser:
Added 'range_check' parameter to allow user to disable the check for overlapping memory regions. The parameter defaults to '1' (enabled) to preserve current functionality. The motivation for this PR is that for certain memory configurations ( large ranges with large interleave ) the check takes an enormous amount of time before truncating the checking algorithm.  Now, the  user can run this check for initial development and then disable it to achieve the best performance.


gvoskuilen:
This will add a lot of spurious merge commits to the history. Can you follow the instructions here: https://github.com/sstsimulator/sst-core/blob/master/CONTRIBUTING.md#forked-development to set up your fork and branch off of the main repo's devel branch? This should eliminate the need for merge commits.

PR #2376, opened 2024-07-17
mem_nic range check enable/disable and test

kpgriesser:
This PR adds a new parameter, range_check, to the MemNIC component. Setting this to 0 will disable the memory region overlap check. 

A test has been added to memHierarchy/tests called testRangeCheck.py.  The test sets the range_check parameter to 0 and should complete very quickly. If the range_check is enabled (or the disable is not working)  the test will stall for a very long time and, presumably, be killed by the auto test scripts.


gvoskuilen (commented):
Code looks good. To add the test to the testsuite, it'll be easiest to change the test filename to match the others, e.g., test_memHA_RangeCheck.py. Then add a function to testcase_memHierarchy_memHA in testsuite_default_memHierarchy_memHA.py. 
```
def test_memHA_RangeCheck:
    self.memHA_Template("RangeCheck")
```
You will also need to commit a reference file to the refFiles dir with a matching name (test_memHA_RangeCheck.out)

gvoskuilen:
I messed up the test filename in my last comment, the filenames should be:
Test: memHierarchy/tests/testRangeCheck.py
Ref: memHierarchy/tests/refFiles/test_memHA_RangeCheck.out

The test logic is in testsuite_default_memHierarchy_memHA.py in the function `memHA_Template`. The template uses the testcase name you pass to it (e.g., "RangeCheck") to identify which files are the test  & ref files.

gvoskuilen (approved):

gvoskuilen:
Make-dist failure probably means the new reference & test files are missing from memHierarchy/Makefile.am so they aren't present in the distribution.

gvoskuilen:
Once testRangeCheck.py is added to Makefile.am, we'll get this retested.

gvoskuilen (approved):

PR #2380, opened 2024-08-01
Ariel MPI

plavin:
This PR adds the ability for the Ariel core model to trace MPI programs. A number of other improvements have been made to Ariel in the process.

The new features are behind a new flag for sst-elements, `--enable-ariel-mpi`. This will cause the programs in `ariel/mpi` to be built.

# Features
- Adds the following options to Ariel:
```
mpimode: Whether to use <mpilauncher> to to launch <launcher> in order to trace MPI-enabled applications.  [0]
mpilauncher: Specify a launcher to be used for MPI executables in conjuction with <launcher>  [MPILAUNCHER_EXECUTABLE]
mpiranks: Number of ranks to be launched by <mpilauncher>. Only <mpitracerank> will be traced by <launcher>.  [1]
mpitracerank: Rank to be traced by <launcher>.  [0]
```
- Adds a directory, `mpi/` which contains `mpilauncher`, and some test programs, `hello.cc`, `reduce.cc`, and `fakepin.cc`.
- Adds `ariel_disable` to the Ariel API, allowing you to stop sending instructions (re-enable output by calling `ariel_enable`)
- Integrates the Ariel API into the build system so that `libarielapi.so` is built by default
- Improves some error messages produced by the Ariel test suite

# Changes
- Previously, whether or not you got a real time or a sim time from `mapped_gettimeofday` and `mapped_clockgettime` depended on whether or not Ariel output was enabled (i.e. whether Ariel is currently sending instructions on the pipe). Now, you will get a simulation time anytime after `ariel_enable` is called. As it was impossible to disable output before, this should not affect any existing programs, but it ensures that if you do use `ariel_disable`, you will have a monotonic clock. 
- NOTE: You should not try to reason about gettimeofday/clockgettime calls made in different `ariel_enable` - `ariel_disable` regions.

# Temporary limitations
- We can only trace a single rank of the MPI program
- SST Core must be compiled without MPI support
- All ranks from the traced program must run on a single node alongside SST
- The Ariel MPI testsuite is not run by default. Run it with `sst-test-elements -w "*mpi_Ariel*"`


plavin:
Errors with rebasing. See #2386.

PR #2381, opened 2024-08-05
Fixes for Vanadis Config Scripts

hughes-c:
Update and fix the kingsley script
Clean the default script


hughes-c:
Fix for https://github.com/sstsimulator/sst-elements/issues/2382

PR #2392, opened 2024-08-23
Ember OTF2 linking and Bcast fix (with an added test)

### LINKED ISSUE #2390: Dynamic linking of libotf2.so is broken
New issue for sst-element
----------------------
1. _Detailed description of problem_
Compiling with OTF2 support works, but trying to actually use OTF2 (via the Ember OTF2 motif) results in `sst: symbol lookup error: libember.so: undefined symbol: OTF2_Reader_Open` because of an issue with the dynamic linking to libotf2.so. There is a workaround: run SST with the `LD_PRELOAD` environment variable set to the path of libotf2.so.

2. _Describe how to reproduce the issue_
Compile with OTF2 support and use the Ember OTF2 motif to read a trace with `addMotif("OTF2 tracePrefix=traces.otf2")`.

4. _What operating system and versions_
TOSS 4.8-2.

6. _What versions of external libraries (MPI, etc.)_
otf2 3.0.3

8. _Provide sha1 of all relevant SST repositories (sst-core, sst-elements, etc.)_
`sst-core c13aff968e9684ceab0f2b50e0887c08091956cd`
`sst-elements 2664955f14ed0ce7d9b0c77d3a18436df5a2c72b`

### LINKED ISSUE #2391: Ember OTF2 motif hangs when attempting to replay a Bcast
New issue for sst-element
----------------------
1. _Detailed description of problem_
Trying to use the Ember OTF2 motif to replay a trace containing an MPI_Bcast will cause the simulation to hang. The motif loads the trace successfully, but then the simulation never completes.

2. _Describe how to reproduce the issue_
Generate the trace of an MPI application containing a Bcast. Then compile sst-elements with OTF2 support and use the Ember OTF2 motif to read that trace with `addMotif("OTF2 tracePrefix=traces.otf2")`.

4. _What operating system and versions_
TOSS 4.8-2.

6. _What versions of external libraries (MPI, etc.)_
otf2 3.0.3
openmpi 4.1.2

8. _Provide sha1 of all relevant SST repositories (sst-core, sst-elements, etc.)_
`sst-core c13aff968e9684ceab0f2b50e0887c08091956cd`
`sst-elements 2664955f14ed0ce7d9b0c77d3a18436df5a2c72b`

freund:
Closes #2390
--------
Fixes the issue with dynamic linking to libotf2.so caused by a minor typo in ember's Makefile.am.

Closes #2391
--------
Fixes the issue of the simulation hanging when the Ember OTF2 motif encounters an MPI_Bcast when replaying a trace. When the OTF2 library reads a trace and encounters a collective operation, it returns the sizes of both the sent and received messages. The motif then replicates this operation in Ember. However, it was issuing a Bcast using the number of sent bytes, which is incorrect because the root rank is the only one sending anything. (For non-root ranks, the number of sent bytes is zero.) It should use the number of received bytes, which is the same for all ranks. I think this discrepancy was causing the simulation to hang because Bcast operations did not match between the different ranks.

Added Ember OTF2 test
--------
I also added a simple test to make sure that the Ember OTF2 motif works correctly. It reads in a small trace (just containing a Bcast across four ranks) and passes as long as the simulation completes within 20 seconds with the correct simulation time.

~~In order to skip this test if sst-elements was not built with OTF2 support, I added a couple lines to ember's Makefile.am to have sst-register add a flag to the configuration file indicating whether it was built with OTF2. The test first checks for that flag (which depends on sstsimulator/sst-core#1121). Please let me know if it's not advisable to pollute the config file in this way!~~ The test will be skipped if sst-elements is found not to have been built with OTF2 support according to the sst-elements config include file.

This test also depends on being able to read in a trace, which comprises 10 small binary files that I've included here. I see that the sst-downloads repo has a release for test support files, so if it's preferable, I'd be happy to adjust my test to download an archive from there instead (it looks like cramSim does something similar). I just wasn't sure what the process of adding files to sst-downloads looked like.

Added support for missing collectives
--------
I added support for Scatter, Allgather, and Alltoall collective operations. They were previously ignored when encountered in an OTF2 trace.

Added parameter for optional compute time
--------
I added a new parameter `addCompute`. It is false by default, but when set to true, the motif will add ember compute events to emulate compute time spent outside of communication. The motif tracks the timestamps of events in the OTF2 trace, beginning with the start of the trace, at the start and end of MPI calls, and at the end of the trace. Using this information along with the time resolution read from the trace's global definitions, the motif can add compute events to the queue that roughly match the elapsed time outside of MPI calls.


berquist:
Marking as WIP until we clean out the Autotester queue.

feldergast (approved):

PR #2406, opened 2024-10-03
Add basic test for simpleElementExample checkpointing

bliu1013:


gvoskuilen:
@bliu1013 Fix the test error by adding testCheckpoint.py to the Makefile.am so that it gets included in the distribution.

gvoskuilen:
The failing test is the one added in this PR: test_basic_checkpoint.

PR #2407, opened 2024-10-04
merlin/bridge: load user subcomponent for networkIF

heyitsanthony:
pymerlin expects to override a SimpleNetwork subcomponent with its own network subcomponent (e.g., LinkControl). When using a Bridge, this subcomponent is ignored and link initialization crashes on SST::Link::sendUntimedData when trying to access a null rtr object. This patch extends the Bridge component to use the user subcomponent if set and if not then fall back to using an anonymous subcomponent as before.

### DIFF
--- a/src/sst/elements/merlin/bridge.h
+++ b/src/sst/elements/merlin/bridge.h
@@ -204,25 +204,29 @@ private:
         dbg.debug(CALL_INFO, 2, 0, "Initializing network interface %d\n", id);
         Nic_t &nic = interfaces[id];
 
-        Params if_params;
+        SubComponentSlotInfo *info = getSubComponentSlotInfo("networkIF");
+        if (info) {
+            int maxSlot = info->getMaxPopulatedSlotNumber();
+            if (maxSlot > 1) {
+                dbg.fatal(CALL_INFO, 1,
+                          "May only specify slots 0 and 1 for 'networkIF'; maximum slot given: %d.\n",
+                          maxSlot);
+            }
+        }
+        if (!info || !info->isPopulated(id)) {
+            Params if_params;
+            if_params.insert("link_bw",params.find<std::string>("network_bw","80GiB/s"));
+            if_params.insert("input_buf_size",params.find<std::string>("network_input_buffer_size", "1KiB"));
+            if_params.insert("output_buf_size",params.find<std::string>("network_output_buffer_size", "1KiB"));
+            if_params.insert("port_name","network" + std::to_string(id));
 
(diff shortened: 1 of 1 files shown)


src/sst/elements/merlin/bridge.h
```diff
@@ -204,25 +204,22 @@ class Bridge : public SST::Component {
...
-
-        if_params.insert("link_bw",params.find<std::string>("network_bw","80GiB/s"));
-        if_params.insert("input_buf_size",params.find<std::string>("network_input_buffer_size", "1KiB"));
-        if_params.insert("output_buf_size",params.find<std::string>("network_output_buffer_size", "1KiB"));
-        if_params.insert("port_name","network" + std::to_string(id));
-
-        nic.nic = loadAnonymousSubComponent<SST::Interfaces::SimpleNetwork>
-            ("merlin.linkcontrol", "networkIF", id,
-             ComponentInfo::SHARE_PORTS | ComponentInfo::INSERT_STATS, if_params, 1 /* vns */);
+        nic.nic = loadUserSubComponent<SST::Interfaces::SimpleNetwork>
```
feldergast, line 207: This should really use one slot name (networkIF), with multiple instances using an index.  In the c++ code, this would use the SubComponentSlotInfo object to load the user subcomponents.  The definition of the class is in baseComponent.h and you would get if by calling:

getSubComponentSlotInfo("networkIF")

This object allows you to query which indices have been loaded with a user subcomponent.

The code would look something like:

        SubComponentSlotInfo* info = getSubComponentSlotInfo("networkIF");
        if ( !info ) {
            // Create anonymous subcomponent
        }
        info->create<SimpleNetwork>(id, ComponentInfo::SHARE_PORTS, 1)

There are also other functions that could be used to do more complete error checking to see if the right number of subcomponents had been added, etc.

feldergast (changes requested):
Please change things to use the SubComponentSlotInfo interface for creating the user subcomponents.

src/sst/elements/merlin/bridge.h
heyitsanthony: OK, thank you for the fast response and detailed explanation on how to properly support this. I updated the code to behave as follows:
* load anonymous subcomponent if no info
* load anonymous component if id is not populated (for half-anonymous case)
* load user component if id is populated
* fatal out if more than two slots are given

I was passing "networkIF1" to `NetworkInterface._instanceNetworkInterfaceBackCompat` beforehand; I changed it to "networkIF" with slot 1 and it works great!

PR #2414, opened 2024-10-30
Vanadis examples

shannong:
Adding initial vanadis example: Single cpu with L1/L2/L3 caches


hughes-c:
@feldergast Will a rebase clean up the commits from devel?

feldergast:
> @feldergast Will a rebase clean up the commits from devel?

Yes, the branch should be rebased against current devel and either force pushed or a new PR created.  Merges from devel into the feature branch should be avoided and rebases should be used instead.  We had a page detailing how this could be done on the website, but I'm not finding it right now.  I'll post a link when I find it.

PR #2422, opened 2024-11-13
merlin/linkcontrol: fix memory leak on insufficient send credits

heyitsanthony:
LinkControl::send allocates a RtrEvent to first compute the number of flits needed to send a request. If there's not enough credits, the method returns early without deleting the RtrEvent object and leaks its memory. This patch deletes the object before returning to fix the leak.


feldergast (approved):
Looks good.  Thanks!

PR #2424, opened 2024-11-21
merlin/hr_router: support user-defined portcontrol subcomponents

heyitsanthony:
For my use-case, I wish to assign bandwidths to local ports different from router-to-router port bandwidths. Although a mechanism exists in hr_router to set heterogeneous port control parameters, it relies on topology logical groups, which is too coarse for assigning properties to local ports (and it's only really used by the dragonfly topology).

This patch checks for user-defined subcomponents on an hr_router's portcontrol slots during initialization and assigns them to the associated port. This is enough to override any port settings by setting the subcomponent on any router after building a topology in pymerlin.

Also includes a small unit-conversion fix for flit_size in portControl that turned up while testing.

### DIFF
--- a/src/sst/elements/merlin/hr_router/hr_router.cc
+++ b/src/sst/elements/merlin/hr_router/hr_router.cc
@@ -245,10 +245,27 @@ hr_router::hr_router(ComponentId_t cid, Params& params) :
     pc_params.insert("oql_track_port", params.find<std::string>("oql_track_port","false"));
     pc_params.insert("oql_track_remote", params.find<std::string>("oql_track_remote","false"));
 
+    SubComponentSlotInfo *pc_info = getSubComponentSlotInfo("portcontrol");
+    if ( pc_info ) {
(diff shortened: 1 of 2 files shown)


feldergast:
I will take a look at this later this week, but for the use case you describe (different bandwidth on local/endpoint ports), there is a simpler mechanism (assuming the local ports are lower bandwidth, which is probably a safe assumption).  Each pair of ports that is connected will "negotiate" the final bandwidth used by the ports by exchanging the set bandwidths and using the lower of the two bandwidths.  So, if you just set the bandwidth of the NICs to be the local bandwidth, those ports will negotiate down to that bandwidth and the routers will use the same bandwidth (again, assuming this bandwidth is lower than the router to router bandwidth).  If you want to control other ports, than something more complex may be needed.

feldergast (commented):
How does this tie in with the Python modules (i.e. what does this look like in a normal configuration file)? Without support in the python module, I'm not sure how a user would be able to use the built-in topology generators.

PR #2427, opened 2024-12-11
Remove calls to deprecated Python test framework functions

berquist:
This removes calls to test functions that were deprecated (https://github.com/sstsimulator/sst-core/pull/1136, https://github.com/sstsimulator/sst-core/pull/1137, https://github.com/sstsimulator/sst-core/pull/1145) and fixes a few other warnings, particularly about unclosed files.

### DIFF
--- a/src/sst/elements/memHierarchy/tests/testsuite_default_memHierarchy_memHSieve.py
+++ b/src/sst/elements/memHierarchy/tests/testsuite_default_memHierarchy_memHSieve.py
@@ -21,7 +21,7 @@ class testcase_memHierarchy_memHSieve(SSTTestCase):
 
 #####
     pin_compiled = testing_is_PIN_Compiled()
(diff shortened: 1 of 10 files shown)


src/sst/elements/memHierarchy/tests/testsuite_default_memHierarchy_memHSieve.py
```diff
@@ -34,7 +34,6 @@ def test_memHSieve(self):
 
     def memHSieve_Template(self, testcase, testtimeout=360):
 
-        pin2defined = testing_is_PIN2_used()
```
berquist, line 37: The else branch for a `if pin3defined:` conditional later on states code exists for Pin 2, but it isn't obvious to me what it does, so I didn't remove it.  If someone (@gvoskuilen ?) could tell me if it's safe to remove, that would be good.

PR #2431, opened 2024-12-18
Use class methods for downloads

jmlapre:
We should be using the standard `setUpClass()` method for time consuming unit test operations such as downloading a file.  This currently works properly with the `sst-unittest` framework in serial mode but does not work in concurrent mode.  I'm marking this as WIP to prevent an automerge from occurring before it gets reviewed.

### DIFF
--- a/src/sst/elements/cramSim/tests/testsuite_default_cramSim.py
+++ b/src/sst/elements/cramSim/tests/testsuite_default_cramSim.py
@@ -9,14 +9,18 @@ import tempfile
 
 
 class testcase_cramSim_Component(SSTTestCase):
+    download_file = None
 
     def setUp(self):
         super(type(self), self).setUp()
+        if self.__class__.download_file is None:
+            self.skipTest("File not downloaded!")
         self._setupcramSimTestFiles()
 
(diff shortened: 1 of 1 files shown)


src/sst/elements/cramSim/tests/testsuite_default_cramSim.py
```diff
@@ -5,14 +5,36 @@
...
+import tempfile
 
 
 class testcase_cramSim_Component(SSTTestCase):
+    download_file = None
 
     def setUp(self):
         super(type(self), self).setUp()
+        if self.__class__.download_file is None:
+            self.skipTest("File not downloaded!")
```
gvoskuilen, line 17: What happens when running the tests concurrently? setUpClass is skipped so these tests will always be skipped?

jmlapre: Yes, that's the current behavior.  The fix is more involved than simply calling `setUpClass()` where `initializeClass()` is called...

jmlapre:
This PR and #2506 are basically the same.  Closing.

PR #2436, opened 2025-01-13
Add latency tracking to Ariel

plavin:
This PR adds latency statistics to `ArielCore`s. It also changes the `ArielCore`s to use a `timeconverter` passed from the `ArielCPU`, instead of relying on default behavior.


src/sst/elements/ariel/arielcpu.h
```diff
@@ -92,6 +92,8 @@ class ArielCPU : public SST::Component {
     SST_ELI_DOCUMENT_STATISTICS(
         { "read_requests",        "Statistic counts number of read requests", "requests", 1},   // Name, Desc, Enable Level
         { "write_requests",       "Statistic counts number of write requests", "requests", 1},
+        { "read_latency",         "Statistic counts latency of read requests", "requests", 1},   // Name, Desc, Enable Level
```
gvoskuilen, line 95: "requests" is the unit of the stat counting, so should be "cycles" I think?

gvoskuilen (approved):

PR #2437, opened 2025-01-14
Cleanup the build a tiny bit

hughes-c:
- Fix a bug in llyr build
- Move common flags in ariel build (make it easier to start testing with clang)
- Remove CUDA build items in ariel make


gvoskuilen (approved):

hughes-c:
AM_LDFLAGS was missing; LDFLAGS seems to be missing from several of the makefiles, which means that user-passed flags aren't being applied (but that's a larger cleanup effort and might break things).

PR #2443, opened 2025-01-21
memH: Add automatic cache flush at shutdown

gvoskuilen:
Added automatic cache flush when saving memory state to file
Fixed some bugs in noninclusive coherence protocols
Updated memory backing to allow separate input/output files
Changed standardCPU to make request stream deterministic regardless of cache state
Updated .gitignore for element python modules, vscode, and test output directory


gvoskuilen:
WIP pending the following updates:
* Fix failing tests
* Change backing output to be binary-only for file and plaintext only for stdout regardless of backing type. 
* Change dump/test functions in backing to printToFile/printToScreen and update memory controller parameter accordingly

plavin (approved):

PR #2454, opened 2025-02-16
Fix iterator type mismatches caused by the std::map comparison functor being left out

leekillough:
This fixes a build error which was detected when compiling sst-elements with the STL debug option `-D_GLIBCXX_DEBUG`:
```
  CXX      nvm_dimm.lo
../../../../../src/sst/elements/messier/nvm_dimm.cc: In member function 'void SST::MessierComponent::NVM_DIMM::schedule_delivery()':
../../../../../src/sst/elements/messier/nvm_dimm.cc:223:31: error: no match for 'operator=' (operand types are 'std::__debug::map<SST::MessierComponent::NVM_Request*, long long int>::iterator' and 'std::__debug::map<SST::MessierComponent::NVM_Request*, long long int, SST::MessierComponent::NVM_DIMM::NVMReqPtrCompare>::iterator')
  223 |     st_1 = ready_at_NVM.begin();
      |                               ^

(similar error for en_1)

  CXX      tlb_unit.lo
../../../../../src/sst/elements/samba/tlb_unit.cc: In member function 'bool TLB::tick(SST::Cycle_t)':
../../../../../src/sst/elements/samba/tlb_unit.cc:231:56: error: no match for 'operator=' (operand types are 'std::__debug::map<SST::MemHierarchy::MemEventBase*, int>::iterator' and 'std::__debug::map<SST::MemHierarchy::MemEventBase*, int, SST::SambaComponent::MemEventPtrCompare>::iterator')
  231 |                   same_st = SAME_MISS[addr/4096].begin();
      |                                                        ^

(similar error for same_en)

../../../../../src/sst/elements/samba/tlb_unit.cc: In member function 'bool TLB::tick(SST::Cycle_t)':
../../../../../src/sst/elements/samba/tlb_unit.cc:360:29: error: no match for 'operator=' (operand types are 'std::__debug::map<SST::MemHierarchy::MemEventBase*, long unsigned int>::iterator' and 'std::__debug::map<SST::MemHierarchy::MemEventBase*, long unsigned int, SST::SambaComponent::MemEventPtrCompare>::iterator')
  360 |         st = ready_by.begin();
      |                             ^

(similar error for en)
```

The iterators are declared without the comparison functor being passed to the `std::map` container in the iterator declaration, causing a type mismatch in the assignments of the iterators.

It may be that  `-D_GLIBCXX_DEBUG`  causes the STL containers to use different default comparison functors, such as debugging wrappers, while the regular defaults match.

Re-declaring the iterators with `auto` ensures that they are the correct type, and fixes this error.

(Some trailing whitespace was removed in the edited files, which is my editor's default behavior.)

### DIFF
--- a/src/sst/elements/messier/nvm_dimm.cc
+++ b/src/sst/elements/messier/nvm_dimm.cc
@@ -219,9 +219,8 @@ bool NVM_DIMM::tick()
 void NVM_DIMM::schedule_delivery()
 {
 
-    std::map<NVM_Request *, long long int>::iterator st_1, en_1;
-    st_1 = ready_at_NVM.begin();
-    en_1 = ready_at_NVM.end();
+    auto st_1 = ready_at_NVM.begin();
+    auto en_1 = ready_at_NVM.end();
 
     while (st_1 != en_1)
     {
--- a/src/sst/elements/samba/tlb_unit.cc
+++ b/src/sst/elements/samba/tlb_unit.cc
@@ -91,13 +91,13 @@ TLB::TLB(ComponentId_t id, int tlb_id, TLB * Next_level, int Level, SST::Params&
 
 
     // === Init Counters
-    
+
 	hits=misses=0;
 
 
     // === Per page-size params =======================================
-    
-    
+
+
     // params
 	size = new int[sizes];
 	assoc = new int[sizes];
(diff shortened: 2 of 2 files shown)


leekillough:
It's likely that the non-debug `std::map<K,V,CMP,ALLOC>::iterator` type does not depend on the comparison functor template argument `CMP`, but in the debug version it does depend on `CMP`, because the debug version performs checks on iterator validity, and the iterator wrapper type depends on all container template arguments.

In any case, there was a mismatch between the `std::map<...>::iterator` types with default- and non-default comparison functors. By using `auto` the iterators are declared with the correct types, even if they depend on the comparison functor.

gvoskuilen (approved):

gvoskuilen (approved):

PR #2455, opened 2025-02-19
Add new element, golem

PlatinumCD:
# Add Golem Element to SST-Elements

## Summary

This PR introduces **Golem**, a new element in `sst-elements`, extending the **Vanadis RoCC Interface**. Golem implements **Matrix-Vector Multiplication (MVM)** through two major subcomponent types:

- **RoCCAnalog** – A RoCC-based interface for handling MVM operations.
- **ComputeArray** – A modular compute array supporting different data types.

### RoCCAnalog
- Implements **RoCC commands** for setting matrices, loading vectors, computing MVM, storing results, and moving vectors between arrays.
- Supports **both integer and floating-point** implementations.
- Interfaces with **Vanadis and StandardMem** for memory operations.

### ComputeArray
- Defines an **abstract base class** for compute arrays.
- Includes specialized implementations:
  - **MVMComputeArray** (Manual MVM in software)
  - **CrossSimComputeArray** (Integrates with CrossSim for analog MVM)
- Supports **float and int64_t** data types.

## New Components
- `RoCCAnalog`
- `RoCCAnalogFloat`
- `RoCCAnalogInt`
- `MVMComputeArray`
- `CrossSimComputeArray`
- `MVMFloatArray`
- `MVMIntArray`
- `CrossSimFloatArray`
- `CrossSimIntArray`

## Dependencies
- Requires **NumPy** for CrossSim-based compute arrays.
- Requires **CrossSim** python library.
- Requires custom [llvm compiler](https://github.com/PlatinumCD/llvm-project/tree/llvm-riscv) for compiling tests.
- Integrates with **Vanadis RoCC** and **StandardMem** for memory interaction.

@bfeinberg will be making additional changes to the Readme. This pull request should be labelled as WIP.


gvoskuilen:
@PlatinumCD The tests are failing because golem requires numpy and I expect they will also fail even with a fix for that because we don't have CrossSim installed. Can you update the compile such that sst-elements has a configure flag with "--with-crossim=<path>" and if Elements is not compiled with crossim, then the golem compile/tests/etc are skipped? I am assuming that numpy is already a dependence of crossim so if you have crossim installed, you should also be able to compile golem? For an example of a conditional compile, look at the memHierarchy backends (such as sst-elements/config/sst_check_ramulator.m4 and sst-elements/src/sst/elements/memHierarchy/Makefile.am).

bfeinberg:
Would it make more sense to add a `--with-numpy=` flag instead because that is that is actually what is being included here? That would allow other elements that want to plug into numpy for whatever reason to reuse the same feature.

gvoskuilen:
I didn't go over all the includes in detail: golem will compile as long as numpy is present, even if CrossSim is not? In that case, yes, it could just be dependent on numpy existing. If yes, then sure we can compile it anytime numpy is found. 

In that case, maybe it makes sense to not have an explicit config flag, always check for numpy, and compile if it's found. Do we also need to disable golem if SST is compiled with an older version of python?

bfeinberg:
After discussion with @gvoskuilen, we're going to add a check for numpy to the configure script. If numpy isn't available we will not compile the Golem models which use numpy (e.g., the CrossSim models).

gvoskuilen (approved):

PR #2461, opened 2025-02-27
Significantly reduce compiler warnings in SST Elements

leekillough:
This fixes compiler warnings which fill build output. Several kinds of warnings were addressed in this PR:
1.  `%llu` and `%llx` were being used to print 64-bit values, when actually the `uint64_t` type used for 64-bit data is `unsigned long`. Although `sizeof(unsigned long) == sizeof(unsigned long long)` in `LP64`, `unsigned long` and `unsigned long long` are considered _distinct_ types, and hence the warnings. The `%llu` and `%llx` strings were replaced with `"%" PRIu64` and `"%" PRIx64`, respectively, to ensure that the format string exactly matches the `uint64_t` arguments.
2. There are a few places where there are true `printf()` and `scanf()` format mismatches. This required changing the format string to use a different format length, to use `PRI*` or `SCN*` macros, or required adding an explicit cast to a `printf()` argument.
3. Security warnings about a `printf()` format not being a constant can be fixed by using the `%s` format and passing the string as a string to be printed rather than a format to be printed.
4.  `(void) !` was added to ignore the return values of functions declared `[[nodiscard]]` or `__attribute__((warn_unused_result))`, removing hundreds of warnings. `(void) !` needs to be used instead of simply `(void)` because in GCC, `(void)` [is insufficient](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425).
5. In `mercury/components/operating_system.cc`, `extern "C"`, if used in a variable declaration (`extern "C" type var;`) rather than a block scope linkage specifier (`extern "C" { }`), cannot _define_ variables. It can only _declare_ them as `extern` and as of `C` linkage, so they can't be defined (initialized) in the same statement. I'm surprised the compiler only gives a warning instead of an error. To _define_ them as `C` linkage variables, they need to be in an `extern "C" { }` block.
6. In `iris/libfabric/prov/sumi/src/sumi_av.cc`, line 200, there is a `snprintf()` which gets a truncation warning, which is not fixed by this PR. But related to it, a buffer overflow can be caused by the size of the string target passed in being `512` instead of `SUMI_MAX_ADDR_LEN`, which is the size of the memory buffer allocated a few lines earlier. A hardcoded value of `512` was used incorrectly, which could cause buffer overflow. The code is still probably wrong, since it may be truncating a `nnn.nnn.nnn` string with too small of a value of `SUMI_MAX_ADDR_LEN`, which gets a warning.
7. In `vanadis/util/vfpreghandler.h`, when a floating-point value is split into two 32-bit integer registers, the code reads past the end of data if the floating-point value is single-precision, because it always copies 8 bytes from the floating-point source to the two 32-bit integer halves. A test was added to be sure the source type is 8 bytes before copying 8 bytes. If it is 4 bytes, then the second 32-bit half will be zero.
8. In `memHierarchy/memoryController.cc`, a filename argument was not passed when printing an error message.
9. In `iris/libfabric/include/linux/osd.h`, there are dozens of warnings about implicit function declarations for `mremap()`. Since `mremap()` is not POSIX, since `osd.h` is versioned for different OSes, and since changing the code may break it for some platforms including the one it is intended to target, these warnings were left intact.


src/sst/elements/vanadis/util/vfpreghandler.h
```diff
@@ -46,6 +46,7 @@ fractureToRegisters(VanadisRegisterFile* regFile, uint16_t r_left, uint16_t r_ri
     std::memcpy((char*)&left_val, value_ptr, sizeof(left_val));
 
     uint32_t right_val = 0;
+    if(sizeof(T) == 8)
```
hughes-c, line 49: Formatting

src/sst/elements/iris/libfabric/prov/sumi/src/sumi_av.cc
```diff
@@ -197,7 +197,7 @@ DIRECT_FN const char *sumi_av_straddr(struct fid_av *av,
     uint32_t rank = ADDR_RANK(*addr_ptr);
     uint16_t cq = ADDR_CQ(*addr_ptr);
     uint16_t rx = ADDR_QUEUE(*addr_ptr);
-    snprintf(ret, 512, "%" PRIu32 ".%" PRIu16 ".%" PRIu16,
+    snprintf(ret, SUMI_MAX_ADDR_LEN, "%" PRIu32 ".%" PRIu16 ".%" PRIu16,
```
hughes-c, line 200: @feldergast Is this correct?

src/sst/elements/memHierarchy/memoryController.cc
```diff
@@ -348,7 +348,7 @@ MemController::MemController(ComponentId_t id, Params &params) : Component(id),
         // Test outfile to find issues before simulation begins
         if ( backing_outfile_ != "" ) {
             auto fp = fopen(backing_outfile_.c_str(),"wb+");
-            sst_assert(fp, CALL_INFO, -1, "%s, ERROR: Unable to open 'backing_out_file'. Is filepath accessible? Filename='%s'\n", getName().c_str());
+            sst_assert(fp, CALL_INFO, -1, "%s, ERROR: Unable to open 'backing_out_file'. Is filepath accessible? Filename='%s'\n", getName().c_str(), backing_outfile_.c_str());
```
hughes-c, line 351: @gvoskuilen Did you change this recently?

hughes-c (changes requested):

src/sst/elements/vanadis/util/vfpreghandler.h
leekillough: I did not use `clang-format` because the changes would be too large, and even if I indented the code in the `if` statement, it would add whitespace changes which distract from the important diffs emphasized by this PR. The diffs were minimized to fix warnings or bugs, but not to reformat.

src/sst/elements/iris/libfabric/prov/sumi/src/sumi_av.cc
leekillough: The hard-coded `512` is definitely wrong, since the `new` a few lines above it uses `SUMI_MAX_ADDR_LEN` which is shorter than `512`. (`snprintf` and the like take the full size of the array and always leave one byte for the `\0` terminator at the end.)

There is still a compiler warning about the output being truncated, since `SUMI_MAX_ADDR_LEN` is not large enough to hold arbitrary integers of the stated format, which is another issue which I am not the best person to address. If it worked before, then this change should not break it.

src/sst/elements/memHierarchy/memoryController.cc
leekillough:
```
e5071a8b80 src/sst/elements/memHierarchy/memoryController.cc (gvoskuilen         2025-02-05 08:48:26 -0700 351)             sst_assert(fp, CALL_INFO, -1, "%s, ERROR: Unable to open 'backing_out_file'. Is filepath accessible? Filename='%s'\n", getName().c_str());

commit e5071a8b80d984379c8512c44ecacd757fb048a1
Author: gvoskuilen <3968723+gvoskuilen@users.noreply.github.com>
Date:   Wed Feb 5 08:48:26 2025 -0700

    memH: Add automatic cache flush at shutdown (#2443)

    memH: Add automatic cache flush at shutdown if saving memory state to file
    * Required a change to standardCPU to make request stream deterministic regardless of cache state
    * Update .gitignore
    * Output memory backing to binary
    * Add option to print memory state to screen (plaintext)
    * Add tests for all backing output/input combos
    * Add test for initializing memory from CPU
```

src/sst/elements/memHierarchy/membackend/backing.h
```diff
@@ -191,15 +191,15 @@ class BackingMalloc : public Backing {
         if (!fp) throw 1;
 
         size_t buffer_size;
-        fread(&buffer_size, sizeof(size_t), 1, fp);
-        fread(&alloc_unit_, sizeof(unsigned int), 1, fp);
-        fread(&shift_, sizeof(unsigned int), 1, fp);
-        fread(&init_, sizeof(bool), 1, fp);
+        (void) !fread(&buffer_size, sizeof(size_t), 1, fp);
```
feldergast, line 194: I understand casting to void to ignore the result, but why the ! ?

leekillough: > I understand casting to void to ignore the result, but why the ! ?

See the original PR comments at top.

src/sst/elements/memHierarchy/memoryController.cc
gvoskuilen: Yes, as @leekillough pointed out, I changed this recently and missed this format arg.

src/sst/elements/iris/libfabric/prov/sumi/src/sumi_av.cc
leekillough: To put it another way, the compiler was smart enough to track the `ret` pointer as being allocated with `new char[SUMI_MAX_ADDR_LEN];`, and it warned that the `snprintf()` would be truncated writing a `nnn.nnn.nnn` string to the `SUMI_MAX_ADDR_LEN`-sized array, even though the size argument passed to `snprintf()` was a hard-coded `512`. With a size of `512` passed to `snprintf()`, it could also cause a buffer overflow, writing past the end of the `SUMI_MAX_ADDR_LEN`-sized array.

```c++
  char* ret = new char[SUMI_MAX_ADDR_LEN];

  snprintf(ret, 512, "%" PRIu32 ".%" PRIu16 ".%" PRIu16,               // WRONG!!!
  snprintf(ret, SUMI_MAX_ADDR_LEN, "%" PRIu32 ".%" PRIu16 ".%" PRIu16, // RIGHT
```

This change changes the `snprintf()` to use the correct size of the array, but it does not remove the truncation warning, which still exists. If the behavior of the old code was to write the full string into `ret` by **lying** to `snprintf()` by telling it that the size was `512` instead of `SUMI_MAX_ADDR_LEN`, and if the `nnn.nnn.nnn` string it generated worked "okay", then this was just luck, and the `snprintf()` was possibly clobbering memory outside of the array which could have any sorts of effects. If the new code, by truncating the result to fit in `SUMI_MAX_ADDR_LEN`, generates a string which is undesirable because it's truncated, then the `SUMI_MAX_ADDR_LEN` value needs to be increased to avoid the truncation. Changing the definition of `SUMI_MAX_ADDR_LEN` is **beyond the intended scope of this PR**.

Old behavior: Buffer overflow + compiler warning about truncated result because the **real** size of the array passed to `snprintf()` is too small for a general result of the specified format, regardless of the claimed length of `512` passed to `snprintf()`

New behavior: Result truncated to fit inside the **true** array size of `SUMI_MAX_ADDR_LEN` (a macro which is used in many places), without changing its definition + same compiler warning about truncated result

```
../../../../../../src/sst/elements/iris/libfabric/prov/sumi/src/sumi_av.cc:200:38: warning: '%u' directive output truncated writing 10 bytes into a region of size 8 [-Wformat-truncation=]
```

leekillough:
@PlatinumCD Please see latest commit https://github.com/sstsimulator/sst-elements/pull/2461/commits/13b6dd6325fce1533cd16ef6f7272781a48831d1 and refer to [comments at the top](https://github.com/sstsimulator/sst-elements/pull/2461#issue-2883782743) explaining the rationale for most of the changes.

This latest commit:
- Removes warnings about `uint64_t` `printf()` `%llx` arguments by using `PRIx64` (see rationale at top of PR)
- Uses uniform initialization syntax `uint64_t{nnn}` to create constants of type `uint64_t` passed to `printf()` (previously they were `int` which does not match format, and using `uint64_t{nnn}` is safer than using `ul` or `ull` suffixes because it is always `uint64_t` and the `{}` syntax produces an compile-time error if `uint64_t` cannot exactly represent the constant, i.e., if it's narrowing)
- Removes warning about assigning string literals to `char*` which should be `const char*`
- Removes warning about `printf()` format security by using `%s` (see rationale at top)

PlatinumCD:
@leekillough It looks good! Thank you

berquist:
In a separate PR there definitely needs to be a (possibly platform-dependent) fix for the `mremap` warnings.  I haven't tried GCC 13.3 yet (I think we have 13.2 on a CI machine) but on 14.2 these are errors.

leekillough:
> In a separate PR there definitely needs to be a (possibly platform-dependent) fix for the `mremap` warnings. I haven't tried GCC 13.3 yet (I think we have 13.2 on a CI machine) but on 14.2 these are errors.

There is a `#define _GNU_SOURCE` commented-out in the Linux `osd.h` which, according to the manpages, needs to be defined if `mremap()` is to be enabled. I did not uncomment this because it's out of the scope of removing warnings.

hughes-c (approved):

PR #2464, opened 2025-03-05
Subcomponent fix for checkpointing

kpgriesser:
This fixes the subcomponent for issue [(sstsimulator/sst-core#1228)](https://github.com/sstsimulator/sst-core/issues/1228). It adds the required constructor and macro for each subcomponent so that the test can be run with checkpointing enabled. I went ahead with this PR so the code could be reviewed as I'm basing other subcomponent checkpointing models on this code.

### DIFF
--- a/src/sst/elements/simpleElementExample/basicSubComponent_subcomponent.h
+++ b/src/sst/elements/simpleElementExample/basicSubComponent_subcomponent.h
@@ -89,7 +89,11 @@ public:
 
     int compute( int num) override;
     std::string compute( std::string comp ) override;
+
+    // serialization
+    basicSubComponentIncrement();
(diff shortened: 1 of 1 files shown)


src/sst/elements/simpleElementExample/basicSubComponent_subcomponent.h
```diff
@@ -89,7 +89,11 @@ class basicSubComponentIncrement : public basicSubComponentAPI {
 
     int compute( int num) override;
     std::string compute( std::string comp ) override;
+
+    // serialization
+    basicSubComponentIncrement();
```
gvoskuilen, line 94: Where is the implementation for these constructors?

gvoskuilen (approved):

PR #2476, opened 2025-03-24
Add TLS to Mercury

jmlapre:
Add TLS support to Mercury.


src/sst/elements/mercury/operating_system/process/app.cc
jpkenny, line 1: IIRC, I added the global.cc/global.h files and commented out all the global stuff in app. Since it's working, can you delete all of the commented out code here?

jpkenny (changes requested):

src/sst/elements/mercury/operating_system/process/app.cc
jmlapre: Sure thing.

jpkenny (approved):
I'm confused by some of the changes that I thought were already there being marked as new, but it all looks correct so let's see what the autotester tells us.

PR #2477, opened 2025-03-24
Fix: set/get Affinity

PlatinumCD:
Each process has an affinity that is now properly read and set from the getaffinity and setaffinity syscalls. Additionally, the getcpu returns the value of the logical core. 

## Important Note
Since context switching is NOT implemented, setting the affinity does not trigger a context switch.

## Steps to implement context switching:
1. Kernel Virtual Memory:
- Map kernel space into virtual memory.
2. Process Control Block (PCB):
- Define and initialize the PCB structure for each process.
3. Scheduler & Ready Queues:
- Design and implement the scheduler.
- Manage ready queues to organize process contexts.
4. Context Saving:
- Save the current thread state.
- Halt hart execution and store context in the PCB.
- Ensure TLB entries are flushed or updated to prevent stale mappings.
5. Context Loading:
- Retrieve the next process context from the ready queue.
- Load the context and resume hart execution.
- Update MMU settings and refresh the TLB as needed.


src/sst/elements/vanadis/os/syscall/getcpu.h
```diff
@@ -27,8 +27,10 @@ class VanadisGetcpuSyscall : public VanadisSyscall {
     VanadisGetcpuSyscall( VanadisNodeOSComponent* os, SST::Link* coreLink, OS::ProcessInfo* process, VanadisSyscallGetxEvent* event )
         : VanadisSyscall( os, coreLink, process, event, "getcpu" ) 
     {
-        m_output->verbose(CALL_INFO, 16, 0, "[syscall-getcpu] cpu=%d\n",process->getCore());
-        setReturnSuccess(process->getCore());
+        m_output->verbose(CALL_INFO, 16, 0, "[syscall-getcpu] cpu=%d hwThread=%d\n",process->getCore(),process->getHwThread());
```
gvoskuilen, line 30: Format specifiers should be `%u` (unsigned).

src/sst/elements/vanadis/os/include/process.h
```diff
@@ -134,8 +138,8 @@ class ProcessInfo {
     }
 
     ProcessInfo( SST::Output* output, std::string checkpointDir,
-        MMU_Lib::MMU* mmu, PhysMemManager* physMemMgr, int node, unsigned pid, VanadisELFInfo* elfInfo, int debug_level, unsigned pageSize , uint64_t coreCount)
-        : m_mmu(mmu), m_physMemMgr(physMemMgr), m_pid(pid), m_pgid(pid), m_tid(pid), m_elfInfo(elfInfo), m_pageSize(pageSize), m_coreCount(coreCount)
+        MMU_Lib::MMU* mmu, PhysMemManager* physMemMgr, int node, unsigned pid, VanadisELFInfo* elfInfo, int debug_level, unsigned pageSize , uint64_t numLogicalCores)
```
hughes-c, line 141: You declared numLogicalCores as uint32_t. It probably doesn't matter much if it's 32b or 64b, but be consistent.

hughes-c (changes requested):

hughes-c:
@Anunalla Is this going to conflict with your work?

Anunalla:
No.

gvoskuilen (approved):

PR #2484, opened 2025-03-26
Fix use_after_move

jmlapre:
We `std::move(f)` on line 230.  According to the spec this leaves `f` in a valid but unspecified state.  On line 235 we call `f` which we should not be doing.


jmlapre:
For clarity, here is the entire `SimTransport::allocateCq()` method:
https://github.com/sstsimulator/sst-elements/blob/32abe24c63e8f3af6e74ba8065af6ba3fecabf9c/src/sst/elements/iris/sumi/sim_transport.cc#L227-L239

jpkenny (approved):

PR #2488, opened 2025-03-31
fix misc warnings

leekillough:
This fixes miscellaneous warnings found by Clang/LLVM using the default build's compiler flags.

@gvoskuilen In `memHierarchy/directoryController.cc`, a useless comparison looks like it should have been an assignment.

@jpkenny In `mask-mpi/mpi_comm/mpi_comm_cart.h`, there was an off-by-1 comparison of an array index. I fixed it and also flagged negative indices.

In `mask-mpi/tests/sendrecv.cc` and `memHierarchy/cacheFactory.cc`, variable-length array warnings were removed by making their dimension variables `const`.  (There are still other cases of VLAs in Elements which are non-`const` in size.)

In `vanadis/lsq/vbasiclsq.h`, redundant parentheses around a comparison were removed.

### DIFF
--- a/src/sst/elements/mask-mpi/mpi_comm/mpi_comm_cart.h
+++ b/src/sst/elements/mask-mpi/mpi_comm/mpi_comm_cart.h
@@ -64,7 +64,7 @@ class MpiCommCart : public MpiComm
   virtual ~MpiCommCart() {}
 
   int dim(int i) const {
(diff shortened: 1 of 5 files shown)


hughes-c (commented):
The memH and vanadis changes look fine. I'm not sure about the mpi change.

gvoskuilen (approved):

PR #2489, opened 2025-03-31
memH: Add ability to read memory during init()

gvoskuilen:
Also test cleanup to begin removing redundant tests
Adds some example configs separate from tests


src/sst/elements/memHierarchy/tests/mhlib.py
plavin, line 1: A `link_params` dict argument that can allow users to set default parameters for every link might be a useful addition that would save having to construct the tuple list to enable debugging. Or just a `link_debug=<n>` param.

gvoskuilen: That would simplify things. I added a "link_params" argument to both the constructor and connect() function. I left the tuple construct too so that there's flexibility to also add params to a subset of the links.

plavin (approved):

Issue #2491, opened 2025-04-03
distclean Inadvertently Removes Files From In-Source Builds


hughes-c:
When not doing out-of-source builds, distclean removes these files from the source directory:
        src/sst/elements/ariel/api/arielapi.c
        src/sst/elements/ariel/api/arielapi.h
        src/sst/elements/ariel/mpi/fakepin.cc
        src/sst/elements/ariel/mpi/mpilauncher.cc

https://github.com/sstsimulator/sst-elements/commit/6ecf9bde3350699455b27105daaf81550cf6c813

hughes-c:
@plavin Is there a reason these files need to be copied into build? They can't be referenced from source?

plavin:
Thanks for fixing that. I had a WIP fix but hadn't pushed it yet. I had issues with out-of-source builds that I was attempting to workaround. Referencing the source is a better solution.

PR #2492, opened 2025-04-04
Fix for api and mpi file deletion when building in-source

hughes-c:
Fix for https://github.com/sstsimulator/sst-elements/issues/2491

### DIFF
--- a/src/sst/elements/ariel/Makefile.am
+++ b/src/sst/elements/ariel/Makefile.am
@@ -247,7 +247,6 @@ install-exec-hook:
 	$(MKDIR_P) $(libexecdir)
 	$(INSTALL) fesimple.so $(libexecdir)/fesimple.so
 	$(MKDIR_P) $(pkglibdir)
-	$(INSTALL) api/libarielapi.so $(pkglibdir)/libarielapi.so
 	$(SST_REGISTER_TOOL) SST_ELEMENT_SOURCE     ariel=$(abs_srcdir)
 	$(SST_REGISTER_TOOL) SST_ELEMENT_TESTS      ariel=$(abs_srcdir)/tests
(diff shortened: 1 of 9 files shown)


plavin:
When building out of source, `ariel/api/libarielapi.so` is missing.

Configure: 
```
./configure --prefix=$SST_INSTALL --with-pin=$INTEL_PIN_DIRECTORY --enable-ariel-mpi
```

Error: 
```
$ make install
[ ... omitted ...]
/usr/bin/install -c api/libarielapi.so /home/prlavin/sst-prs/install/lib/sst-elements-library/libarielapi.so
/usr/bin/install: cannot stat 'api/libarielapi.so': No such file or directory

```

hughes-c:
I think that is being caused by the ariel makefile explicitly trying to copy it. I think that automake and libtool should handle it.

plavin:
I'm working on a couple issues with this:
1. When running `make` twice on an in-source build, `cp` will refuse to copy `mpilauncher` as the source and destination are the same.
2. The io and mpi tests fail on out-of-source builds. (`sst-test-elements -w "*Ariel*"`)

hughes-c:
Is it because the test binaries aren't being built due to paths or something else?

plavin:
The binaries exist in the build directory, so it is a paths issue.

plavin:
Actually, libarielapi.so is missing from the build directory - it is only in the source directory

plavin:
The MPI tests are still broken so this remains a WIP

plavin:
This should be ready to go now. MPI tests work. The previous two commits removed the copy files hack as well as a path hack from the MPI tests.

gvoskuilen (approved):

plavin (approved):

PR #2493, opened 2025-04-04
Developed a new Ariel Frontend that use EP Analytics tools to collect the address stream

accauble:
# New Ariel Frontend

This PR develops a new ariel frontend so that users can have an alternate instrumentation tool. EP Analytics has two binary instrumentation tool kits, PEBIL (x86) and EPAX (arm64), which can be used with the frontend.

## Changes to Ariel

The changes to Ariel were pretty localized. I created a new class, `ArielFrontendCommon`, which handles much of the code shared by the pin and epa frontends. I kept it out of the `ArielFrontend` class to allow other frontends to be added quickly without having to implement what is required by `ArielFrontendCommon`. 

The bulk of the changes to the pin frontend are cosmetic (i.e. code moved to the `ArielFrontendCommon` or just moved around the file).

There are some minor changes to the pintool:
* If an instruction does a read and a write, it gets instrumented with the `WriteInstructionReadWrite` function (before it was just instrumenting it as a read)
* Scatter/gather instructions are now ignored because these are handled differently by PIN and so the current instrumentation instructions break
* Call/return instructions are now ignored only because users are generally less interested in these types of instructions. @plavin should confirm that these should actually be ignored. 

There is a small change to `ArielCPU` to be able to use different frontends instead of automatically using the pin frontend. The pin frontend is still default but can be specified with "ariel.frontend.pin". The new EP Analytics frontend can be specified with "ariel.frontend.epa".

## Changes to Use

There are no changes to using the pin-based ariel frontend.

Pin instruments the application while the application is running. With EPA tools, the application is instrumented in a separate step. The instrumented binary is usually named `<APP_NAME>.arielinst`. Thus, to use the EPA frontend, users will need to source the EPA tool to put it in their environment, instrument the binary with an EPA tool (instructions to come later) and update their SST scripts with the following parameters:

``` 
     "frontend"      : "ariel.frontend.epa",
     "executable"  : "./<APP_NAME>.arielinst"
```

## Changes to Build

`--with-pin` is still required to build the pin frontend. However, since the epa frontend does not require an EPA tool, it can be built even if the EPA tool is not in the environment.

## Changes to Tests

The tests were redone to allow configuration of the frontend. Some reference files had to be renamed to include "pin" but they were otherwise untouched. All the tests were renamed to end with "_pin". An "epa" version of each test was created that uses the epa frontend. These are skipped if an EPA tool is not sourced.


src/sst/elements/ariel/arielcpu.cc
```diff
@@ -142,8 +142,10 @@ ArielCPU::ArielCPU(ComponentId_t id, Params& params) :
 
     frontend = loadUserSubComponent<ArielFrontend>("frontend", ComponentInfo::SHARE_NONE, core_count, maxCoreQueueLen, memmgr->getDefaultPool());
     if (!frontend) {
+        std::string frontendType= params.find<std::string>("frontend", "ariel.frontend.pin");
```
plavin, line 145: The `frontend` parameter is not in`SST_ELI_DOCUMENT_PARAMS` in `arielcpu.h`

plavin (approved):

gvoskuilen:
This is failing because the names of test reference files have been changed and make-dist is still trying to copy in the old ones. Update the `EXTRA_DIST` list in Makefile.am (and also add the new epa ref files).

accauble:
> This is failing because the names of test reference files have been changed and make-dist is still trying to copy in the old ones. Update the `EXTRA_DIST` list in Makefile.am (and also add the new epa ref files).

Thank you so much for the tip, @gvoskuilen! I have made those changes.

plavin:
@gvoskuilen Looks like everything passed. I assume the final check is a maintainer inspection?

PR #2495, opened 2025-04-07
Add dummy hw_probe syscall for rv64

hughes-c:
We should probably add some of this functionality for real at some point:
https://www.kernel.org/doc/html/next/riscv/hwprobe.html


hughes-c:
More changes that I anticipated due to automatic removal of trailing whitespace...

Focus on the HWPROBE additions.

gvoskuilen (approved):
Looks good. We will fix the printf in a future PR to fix all of vanadis' use of printf instead of sst.output.

Issue #2497, opened 2025-04-09
sst-elements autoconf script has trailing whitespace


jleidel:
I know this is a nit, but its nice to keep things tidy.  When executing autoconf on the `devel` branch on 2025.04.09, I see the following warning: 

```
src/sst/elements/memHierarchy/Makefile.am:132: warning: whitespace following trailing backslash
```

I verified that there is indeed whitespace on this line.  I can submit a PR shortly.

berquist:
I can do one better and add the pre-commit hook both here and in core that universally forbids trailing whitespace.

PR #2499, opened 2025-04-09
Remove existing and prevent future trailing whitespace

berquist:
Closes https://github.com/sstsimulator/sst-elements/issues/2497

This ignores any file that looks like a patch or a reference output for a test.

A common concern is that applying these changes affects `git blame`.  The solution is to use the [`.git-blame-ignore-revs`](https://gist.github.com/kateinoigakukun/b0bc920e587851bfffa98b9e279175f2) file.  This is automatically picked up by GitHub but, like all other things with Git, requires some setup (described in the link above) to work locally.  It's also useful to add this to one's `.gitconfig` to specially mark lines that appeared during an ignored commit, such as when a reformatting turned one line into multiple.  (The original or parent line will still show the original proper commit.)
```
[blame]
    coloring = repeatedLines
    markIgnoredLines = true
    markUnblamableLines = true
```

Using this will be especially important for core when applying the updated `clang-format`.

However, because we squash commits, additions to this file need to happen after the PR that makes formatting changes, which is why it isn't added here.


leekillough:
Another way to encourage styles and prevent trailing whitespace, is to create editor configuration files. I don't use `vim`, but here is an `emacs` `.dir-locals.el` file which can be placed at the top directory of a source tree for which it will apply:
```elisp
(
 ;; Makefiles require tabs and are almost always width 8
 (makefile-mode . (
                   (indent-tabs-mode . t)
                   (tab-width . 8)
                   )
                )

 ;; C formatting
 (c-mode . (
            (c-file-style . "k&r")
            (c-basic-offset . 4)
            (comment-start . "/* ")
            (comment-end . " */")
            )
         )

 ;; C++ formatting
 (c++-mode . (
            (c-file-style . "k&r")
            (c-basic-offset . 4)
            (comment-start . "// ")
            (comment-end . "")
            )
         )

 ;; Default formatting for all source files not overriden above
 (prog-mode . (
                (indent-tabs-mode . nil)
                (require-final-newline . t)
                (sentence-end-double-space . nil)
                (eval add-hook `before-save-hook `delete-trailing-whitespace)
               )
            )
)
```
The last `prog-mode` setting prevents tab indentation, adds a newline if it's missing at EOF, removes typewriter-era double spaces after sentences (e.g. when using `M-x` `fill-paragraph`), and deletes trailing whitespace when saving files. 

With these editor settings, I've made a bunch of PRs where whitespace changes were numerous, and I've had to use `git add -p` to ignore them since they were not pertinent to my PR.

berquist:
> Another way to encourage styles and prevent trailing whitespace, is to create editor configuration files. I don't use `vim`, but here is an `emacs` `.dir-locals.el` file which can be placed at the top directory of a source tree for which it will apply:

The reason this doesn't work is that, even if everyone used <insert editor/IDE here>, it is still possible to violate these settings and commit non-conforming changes to version control.  By making it a part of CI, that's impossible.

The downsides are that:
- If you want to run this locally, installing the Python package is required. (not hard)
- If you want your editor/IDE to automatically apply these rules, you need to provide additional configuration for every editor/IDE.  But you would need to do that either way, and the hooks automatically pick up any existing configuration.

The extension of what you show is https://editorconfig.org/, which covers some basics and is editor/IDE agnostic.  That does mean one needs to install `editorconfig-mode`, but it will automatically set `indent-tabs-mode`, `tab-width`, etc. on a per-major-mode basis.  What is still missing is more complicated integrations, such as which `clang-format` mode to install and what settings it needs aside from the tool itself.  Then there is no way around `.dir-locals.el` or equivalent, but again this isn't a requirement.

---

The summary is that enforcing requirements should be as automated as possible so 1. version control history remains clean and 2. developers don't need to become bogged down in configuration unless they want to.

leekillough:
> The reason this doesn't work is that, even if everyone used <insert editor/IDE here>, it is still possible to violate these settings and commit non-conforming changes to version control. By making it a part of CI, that's impossible.

I am not strongly for or against CI enforcement of formatting, but I am for solutions which make it automatic during ordinary editing and commits, which project-local editor configuration files and automatic reformatting during `pre-commit` scripts can help with.

In [some projects](https://github.com/tactcomplabs/rev/blob/1e145cb776f44d9c5aa2b4c3c72b4930057e361c/.githooks/pre-commit), I have a `pre-commit` hook which automatically removes trailing whitespace at commit time -- this is easier and less hassle than requiring someone to edit their file because it violates a `pre-commit` or CI check, and for text files, it is almost always safe (`git grep -Iqe . -- <file>` is an easy way to test if a file is considered text, and then `sed -e 's/[[:space:]]*$//'` can remove trailing whitespace).

hughes-c:
@berquist  Are you happy with the way this is implemented in this PR? The conversation with @leekillough is resolved?

hughes-c (approved):
Lots of changes but almost all white space removal. Can be merged pending resolution of implementation conversation.

berquist:
> @berquist Are you happy with the way this is implemented in this PR? The conversation with @leekillough is resolved?

I am and yes; I think we should consider adding something like Lee uses in the future, but it requires more thought and is not a substitute for the implementation here.

leekillough:
> > @berquist Are you happy with the way this is implemented in this PR? The conversation with @leekillough is resolved?
> 
> I am and yes; I think we should consider adding something like Lee uses in the future, but it requires more thought and is not a substitute for the implementation here.

I am willing to create a `.dir-locals.el` for Emacs which will remove trailing whitespace on save, avoid tab characters in `prog-mode` but not in `makefile-mode`, use the 4-space `c-basic-offset`, etc. Vim users can probably do something similar. I don't know enough about VS Code.

A `pre-commit` hook can also remove trailing whitespace on text files at commit time (detectable with `git grep -I` with certain project-specific filename pattern exclusions, etc.).

It is to _supplement_, not replace, this PR's changes, so that everything "just works" the first time instead of requiring the user to manually fix it with `clang-format`, `sed`, or whatever tools they use every time it rejects a PR because of unwanted trailing whitespace, etc.

PR #2502, opened 2025-04-11
Disable long running tests for non-Nightly tests

jmlapre:
Including:
* EmberSweep
* miranda_randomgen
* vanadis_short_tests

This cuts the running time of `sst-test-elements` by ~20 minutes.

The tests are still there and will run during Nightly testing.

Depends on https://github.com/sstsimulator/sst-core/pull/1290


jmlapre:
Make no mistake: this approach is definitely using a shotgun to fix the problem.  But at least we are exercising the code and the full set of tests still executes on Nightly builds.

gvoskuilen (approved):
Looks good as a first cut. We can reassess individual tests if we start getting a lot of nightly failures.

PR #2503, opened 2025-04-21
Update ser& to SER_SST for all elements except memH

feldergast:


src/sst/elements/mask-mpi/mpi_message.cc
jpkenny, line 1: Double parentheses

src/sst/elements/mercury/operating_system/launch/app_launch_request.h
jpkenny, line 1: Double parentheses

src/sst/elements/thornhill/memoryHeapEvent.h
```diff
@@ -41,10 +41,10 @@ class MemoryHeapEvent : public SST::Event {
 
     void serialize_order(SST::Core::Serialization::serializer &ser)  override {
         Event::serialize_order(ser);
-		ser & type;
-        ser & key;
-        ser & length;
-        ser & addr;
+		SST_SER(type);
```
jpkenny, line 44: Fix indentation

jpkenny (changes requested):
Double parentheses should be fixed, and might as well fix the couple weird indentations at the same time.

src/sst/elements/mask-mpi/mpi_message.cc
jpkenny: "rc_rank"

jpkenny (changes requested):
Looks like one more regex boo-boo.
src_rank turned into rc_rank somehow

hughes-c (approved):
It would be better to include the memH update in the same PR.

PR #2506, opened 2025-04-22
Update cramSim tests

jmlapre:
Now they all use the same download file or skip the tests if the file was not able to be downloaded.

### DIFF
--- a/src/sst/elements/cramSim/tests/testsuite_default_cramSim.py
+++ b/src/sst/elements/cramSim/tests/testsuite_default_cramSim.py
@@ -1,13 +1,17 @@
 # -*- coding: utf-8 -*-
 
+import os
+import shutil
+import tempfile
+
 from sst_unittest import *
 from sst_unittest_support import *
 
-import os
-import shutil
-
 
 class testcase_cramSim_Component(SSTTestCase):
+
+    tmp_dir = None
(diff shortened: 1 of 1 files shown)


src/sst/elements/cramSim/tests/testsuite_default_cramSim.py
```diff
@@ -112,6 +132,9 @@ def cramSim_test_template(self, testcase):
 #####
 
     def _setupcramSimTestFiles(self):
+        if not self.__class__.tmp_file:
+            # if tmp_file is None setUpClass() never ran
+            self.skipTest(f"{self} cannot be run concurrently")
```
gvoskuilen, line 137: Is setUpClass skipped when running concurrently? Or may be run multiple times?

jmlapre: It's skipped... I believe it has something to do with testtools but I haven't had a chance to dig deep enough yet.

gvoskuilen: OK. We may want to revisit how we do one time setup in general since it'd be nice to support concurrent tests again. Looks fine for now though.

gvoskuilen (approved):

PR #2509, opened 2025-04-24
MemH: Checkpoint support for some memH components

gvoskuilen:
With this PR, checkpoint works for memController (no backing), simpleMem membackend, bus, standardcpu and standardinterface
Also addresses some compiler warnings/deprecations


nab880 (approved):

src/sst/elements/memHierarchy/membackend/backing.h
```diff
@@ -175,12 +182,82 @@ class BackingMMAP : public Backing {
         SST_SER(buffer_);
         SST_SER(size_);
         SST_SER(offset_);
+        SST_SER(mmapfile_);
```
hughes-c, line 185: We should probably think about how to do this if we need something that's more performant or in the situation where you have limited space.

gvoskuilen: As is, that line just serializes the filename. In this PR, checkpoint won't work yet for backing stores. We need to add an sst-core API to add files to a checkpoint and locate them on restart so that we can copy the actual file into the checkpoint.

PR #2516, opened 2025-05-06
Fix scanf() format mismatch; move side effects outside of assert()

leekillough:
There was a format mismatch in `scanf()` for `mode_t`, which is an `int` but was treated as a `short`.

```Diff
-        assert( 1 == fscanf(fp,"mode: %hd\n", &mode) );
+        f = fscanf(fp,"mode: %d\n", &mode);
+        assert( 1 == f );
```

`<assert.h>` and `<stdint.h>` were not `#include`d.

Also, `assert()` conditions should not have side effects because `assert()` can be disabled with `-DNDEBUG`.

@mjleven

### DIFF
--- a/src/sst/elements/vanadis/os/include/fdTable.h
+++ b/src/sst/elements/vanadis/os/include/fdTable.h
@@ -16,6 +16,8 @@
 #ifndef _H_VANADIS_NODE_OS_INCLUDE_FD_TABLE
 #define _H_VANADIS_NODE_OS_INCLUDE_FD_TABLE
 
+#include <assert.h>
+#include <stdint.h>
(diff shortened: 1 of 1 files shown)


leekillough:
Also note: if `mode_t` is `int32_t` on some platforms and `int16_t` on others, then the code should be changed to:
```Diff
-        assert( 1 == fscanf(fp,"mode: %hd\n", &mode) );
+        int temp_mode;
+        f = fscanf(fp,"mode: %d\n", &temp_mode);
+        mode = temp_mode;
+        assert( 1 == f );

feldergast (approved):

PR #2517, opened 2025-05-06
Fix missing #includes

leekillough:
Similar to https://github.com/sstsimulator/sst-core/pull/1308 with script run on Elements source.


feldergast (approved):

feldergast:
This is failing with the following error:

```
20_13-47-58/local/sst-core/include -I../../../../src -std=c++17 -g -O2 -MT ctrlMsgProcessQueuesState.lo -MD -MP -MF .deps/ctrlMsgProcessQueuesState.Tpo -c ctrlMsgProcessQueuesState.cc  -fPIC -DPIC -o .libs/ctrlMsgProcessQueuesState.o
14:01:36 In file included from /usr/include/features.h:428,
14:01:36                  from /usr/include/c++/8/x86_64-redhat-linux/bits/os_defines.h:39,
14:01:36                  from /usr/include/c++/8/x86_64-redhat-linux/bits/c++config.h:2458,
14:01:36                  from /usr/include/c++/8/cstddef:49,
14:01:36                  from ctrlMsg.cc:16:
14:01:36 /usr/include/assert.h:64:1: error: expected unqualified-id before string constant
14:01:36  __BEGIN_DECLS
14:01:36  ^~~~~~~~~~~~~
```

Adding in the extra includes may be creating a circular dependance where a #define isn't getting set.  We're going to hold off on this PR until after the release.  Also, let's plan to split it up into one PR per element (or small set of elements) after the release. Closing this version of the PR.

leekillough:
It could also have to do with the canonical ordering of `#include`s causing this script to reorder `#include`s. In SST Core, we usually have two types of `#include`s:
1. System header `#include<>`s, which, with rare exceptions, do not have `.h` between `<>` (C++ headers starting with `c` are preferred to C headers ending in `.h`).
2. Non-system header `#include ""`s, which usually have `.h` in the filename.

To prevent `clang-format` errors enforced on SST Core, we sort contiguous blocks of system header `#include<>`s.

A more complicated sorting mechanism may need to be developed, which may need to list a set of headers to be placed in certain categories, or we may need to add an option to disable sorting of the `#include`s.

PR #2531, opened 2025-05-28
Mercury: overhaul library/app loading and eliminate direct dlopen usage

jpkenny:
- Leaving a couple commits to preserve the work that @nab880 did to trigger loading across simulator ranks by inserting dummy/loader/holder components into libraries/apps.
- This PR significantly overhauls loading libraries/apps through the use of requireLibrary() in SST-Core. This is a much simpler and cleaner, and therefore preferred approach.
- An unanticipated downside of the dummy/loader/holder insertion approach is that the src2src compiler in sst-hgcc fails on the modified code since core output uses thread local variables.


src/sst/elements/mercury/libraries/system/system_library.h
jmlapre, line 1: I'm a little confused about this change...

jpkenny: libsystemlibrary.so is the library name, so "systemlibrary" is correct here.
Previously we were relying on the fact that we dlopened the library ourselves, so core didn't know/care where it actually came from (it didn't need to find it).

jmlapre (approved):

PR #2539, opened 2025-06-30
fix format specifier to remove warning

leekillough:


leekillough:
Although `unsigned long` and `unsigned long long` are byte-for-byte identical in LP64 ABIs, they are considered separate types, and hence the warning (on some compilers) if `%llu` is used with an `unsigned long` argument. I used `PRIu64` because that's the portable way of specifying the format for `uint64_t`.

PR #2540, opened 2025-07-01
MemH checkpoint and code clean up

gvoskuilen:
Three major sets of changes in this PR:
1. Checkpoint support for caches and directory
2. New tests for incoherent caches
3. Code cleanup/organization, mainly moving constructor bodies to .cc files and changing variables to be `_` separated instead of camel case

### DIFF
--- a/src/sst/elements/memHierarchy/coherencemgr/Incoherent.cc
+++ b/src/sst/elements/memHierarchy/coherencemgr/Incoherent.cc
@@ -224,8 +224,8 @@ bool Incoherent::handleGetS(MemEvent * event, bool in_mshr) {
     }
 
     if (mem_h_is_debug_addr(addr) && line) {
-        event_debuginfo_.newst = line->getState();
-        event_debuginfo_.verboseline = line->getString();
+        event_debuginfo_.new_state = line->getState();
+        event_debuginfo_.verbose_line = line->getString();
     }
 
     return true;
@@ -295,8 +295,8 @@ bool Incoherent::handleGetX(MemEvent * event, bool in_mshr) {
         sendNACK(event);
 
     if (mem_h_is_debug_addr(addr) && line) {
-        event_debuginfo_.newst = line->getState();
-        event_debuginfo_.verboseline = line->getString();
+        event_debuginfo_.new_state = line->getState();
+        event_debuginfo_.verbose_line = line->getString();
     }
 
     return true;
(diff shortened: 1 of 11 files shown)


plavin (commented):
A few things I noticed:

- backing.h
	- Are there overheads with calling mmap twice in BackingMMap? The logic could probably be reworked to only call it once. I'm not sure how expensive it is though.
	- The BackingMMap setter and getter that accept vectors don't use the offset_. Is that intended?
	- Is BackingMMap::printToScreen tested? The offset logic is tricky
		- If `size_` is less than `offset_`, will anything be printed?
	- BackingMMap::printToFile unimplemented 
- mshr.h
	-  When is `= default` required? I noticed MSHREntry() and MSHRRegister() don't use it. 
- lineTypes.h
	- Potentially rename `allocated()` -> `isAllocated()`? Not sure if we have a style guide for that.
- coherenceController.h
	- A couple non underscored variable names :`hasThr`, `verboseline`
- CoherenceController::notifyListenerOfAccess
	- Potentially rename `result` -> `result_type`

gvoskuilen:
> A few things I noticed:
> 
>     * backing.h
>       
>       * Are there overheads with calling mmap twice in BackingMMap? The logic could probably be reworked to only call it once. I'm not sure how expensive it is though.
>       * The BackingMMap setter and getter that accept vectors don't use the offset_. Is that intended?
>       * Is BackingMMap::printToScreen tested? The offset logic is tricky
>         
>         * If `size_` is less than `offset_`, will anything be printed?
>       * BackingMMap::printToFile unimplemented
> 
>     * mshr.h
>       
>       * When is `= default` required? I noticed MSHREntry() and MSHRRegister() don't use it.
> 
>     * lineTypes.h
>       
>       * Potentially rename `allocated()` -> `isAllocated()`? Not sure if we have a style guide for that.
> 
>     * coherenceController.h
>       
>       * A couple non underscored variable names :`hasThr`, `verboseline`
> 
>     * CoherenceController::notifyListenerOfAccess
>       
>       * Potentially rename `result` -> `result_type`

backing.h:
* mmap is called once to open the mmap'd backing store and the second call only happens if there is a separate initialization file
* I think there's an issue with offset_ here but it's part of a bigger problem in both backing and memoryController. Both places do address conversion from global to local. I will create a follow-up PR for this.

mshr.h- I started switching a number of the otherwise empty constructors to "default", but there's no real practical difference.

I made the rest of the changes (lineTypes, coherenceController)

plavin (approved):

PR #2546, opened 2025-07-15
mercury: add support for app arguments

jmlapre:
Also add some sample usage to the ostest.py example.


jpkenny:
@jmlapre ostest.cc should verify that the main function is actually getting the right argc/argv passed to it (is it?); that will require a new golden output

PR #2550, opened 2025-07-23
mask-mpi: pass parameters to halo3d

jmlapre:
This should ensure that parameter passing continues to work into the future wrt pymercury.  Otherwise, test_halo3d26 will fail.


jpkenny (approved):
LGTM

PR #2552, opened 2025-07-28
mercury: reduce nodes down to one

jmlapre:
This will reduce false negatives in testing and the functionality is replicated elsewhere.


jpkenny (approved):
LGTM

PR #2563, opened 2025-08-25
Performance optimizations in MemH and Vanadis

gvoskuilen:
memH: Optimize performance during init() by adding thread local caches to hold results of expensive computations
vanadis: 
* Fix LSQ statistics so that base and child classes do not re-implement the same stats.
* Fix register stack to return correct info from unused/full/empty functions (affects reg occupancy stats)
* Fix over-counting of register occupancy stats in vanadis.cc when HW threads exist
* Add some comments
* Fixes to stats caused updates to all test ref files

### DIFF
--- a/src/sst/elements/golem/tests/small/mvm_float_array/multi_array/riscv64/sst.stdout.gold
+++ b/src/sst/elements/golem/tests/small/mvm_float_array/multi_array/riscv64/sst.stdout.gold
@@ -13,12 +13,12 @@ all process have exited
  node0.cpu0:decoder.uop_delayed_rob_full.1 : Accumulator : Sum.u64 = 7548444; SumSQ.u64 = 7548444; Count.u64 = 7548444; Min.u64 = 1; Max.u64 = 1; 
  node0.cpu0:lsq.loads_issued.1 : Accumulator : Sum.u64 = 50802; SumSQ.u64 = 50802; Count.u64 = 50802; Min.u64 = 1; Max.u64 = 1; 
  node0.cpu0:lsq.stores_issued.1 : Accumulator : Sum.u64 = 24383; SumSQ.u64 = 24383; Count.u64 = 24383; Min.u64 = 1; Max.u64 = 1; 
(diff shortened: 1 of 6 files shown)


src/sst/elements/memHierarchy/memNICBase.h
```diff
@@ -407,19 +428,34 @@ class MemNICBase : public MemLinkBase {
...
+                    MemRegion ep = *infoset.begin();
+                    for (auto et = std::next(infoset.begin()); et != infoset.end(); et++) {
+                        if (!ep.merge(*et)) {
+                            merged.insert(ep);
+                            ep = *et;
+                        }
+                    }
+                    merged.insert(ep);
+
+                    // Filter the destEndpointInfo regions by the reachable endpoints behind them
```
hughes-c, line 454: You don't want to warn the user when this happens?

src/sst/elements/memHierarchy/memTypes.h
```diff
@@ -270,15 +270,37 @@ class MemRegion {
     // TODO clean this up to something more succinct
     // We need to compute the set intersection of this MemRegion with MemRegion 'o'
     // The intersection may not be describable as a single MemRegion, so a set is returned
+    //
```
hughes-c, line 273: If this only happens on setup(), is the performance gain large enough that it warrants violating SST policy?

src/sst/elements/vanadis/inst/regstack.h
hughes-c, line 1: I don't like mixing size_t and uintxx_t but that's me. ^-^

hughes-c (approved):

src/sst/elements/memHierarchy/memTypes.h
gvoskuilen: In my moderately complex testcase, adding this saved over 7min. Of course, how much that matters depends on how long the run part of the sim is. For short simulations, it matters a lot.

PR #2565, opened 2025-08-29
Fix compile error in memNICBase.h and ensure empty arrays are respected in MemNIC

mgoldstein322:
compile error when printing endpoints in memNICBase in debug mode fixed
add check to make sure empty sources/destinations are kept empty when MemNIC receives an empty array as a parameter


mgoldstein322:
Worked on Friday and now it doesn't... Need to evaluate

PR #2568, opened 2025-09-03
Vanadis: Add dummy rseq syscall for newer llvm compilers

gvoskuilen:
This fix enabled using LLVM 22.0 to compile apps for Vanadis.


plavin (approved):
LGTM

PR #2571, opened 2025-09-08
Fix trafficgen

ziyuezzy:
In the original sst-merlin code, there has been a very old bug for the uniform target generator. This commit fixes this bug and updates the corresponding implementation in offered_load.cc.

Some explanations
In the original sst-merlin code, the following lines create a range that does not correctly cover all possible peers. This is because the first input arg of SSTUniformDistribution is the number of bins, which need to equal to (max-min+1) here.

From [`uniform.h`](https://github.com/sstsimulator/sst-elements/blob/08c5200e95b2acf2ab79345a597d8cb2ff46ee67/src/sst/elements/merlin/target_generator/uniform.h#L59-L70):

```cpp
UniformDist(ComponentId_t cid, Params &params, int id, int num_peers) :
        TargetGenerator(cid)
    {
        min = params.find<int>("min",0);
        max = params.find<int>("max",num_peers - 1);

        gen = new SST::RNG::MersenneRNG(id);

        int dist_size = std::max(1, max-min);
        dist = new SSTUniformDistribution(dist_size, gen);

    }
```

This has been fixed in this commit, and correspondingly in offered_load.cc, it has been made sure that an Endpoint does not generate packet towards itself. Besides, some link naming fixes are also included to clear the warnings.
Demonstrating the bug and the fix
This bug is easy to reproduce with a 4x4 mesh topology, for example with the following config file:
[mesh_bug_reproduce.py](https://github.com/user-attachments/files/22214957/mesh_bug_reproduce.py)

Before this fix, the result shows that the last Endpoint receives no packets:

image

After this fix:
image

I started with the master branch of sst-element-src... but since it is only ahead of the devel branch with merges, I hope this is ok?

### DIFF
--- a/src/sst/elements/merlin/offeredload/offered_load.cc
+++ b/src/sst/elements/merlin/offeredload/offered_load.cc
@@ -323,7 +323,13 @@ void
 OfferedLoad::progress_messages(SimTime_t current_time) {
     while ( (next_time <= current_time) && link_if->spaceToSend(0,packet_size) ) {
         offered_load_event* ev = new offered_load_event(next_time);
-        SimpleNetwork::Request* req = new SimpleNetwork::Request(packetDestGen->getNextValue(), id, packet_size, true, true, ev);
+        int dest_id=id;
+        while (dest_id == id)
+        {   // Make sure the endpoint is not sending to itself
+            dest_id = packetDestGen->getNextValue();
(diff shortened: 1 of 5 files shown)


ziyuezzy:
sorry, need to start from devel branch instead of master

PR #2572, opened 2025-09-08
fixing uniform targetgen in sst-merlin

ziyuezzy:
-------------------------------------------
In the original sst-merlin code, there has been a very old bug for the uniform target generator. This commit fixes this bug and updates the corresponding implementation in offered_load.cc.

-------------------------------------------
Some explanations
-------------------------------------------
In the original sst-merlin code, the following lines create a range that does not correctly cover all possible peers. This is because the first input arg of SSTUniformDistribution is the number of bins, which need to equal to (max-min+1) here.

From [`uniform.h`](https://github.com/sstsimulator/sst-elements/blob/08c5200e95b2acf2ab79345a597d8cb2ff46ee67/src/sst/elements/merlin/target_generator/uniform.h#L59-L70):

```cpp
UniformDist(ComponentId_t cid, Params &params, int id, int num_peers) :
        TargetGenerator(cid)
    {
        min = params.find<int>("min",0);
        max = params.find<int>("max",num_peers - 1);

        gen = new SST::RNG::MersenneRNG(id);

        int dist_size = std::max(1, max-min);
        dist = new SSTUniformDistribution(dist_size, gen);

    }
```
-------------------------------------------
List of changes
-------------------------------------------

- The bug described above has been fixed in this commit, 
- And correspondingly in offered_load.cc, it has been made sure that an Endpoint does not generate packet towards itself. 
- Some link naming- The bug described above has been fixed in this commit, 
- Besides, although the ''trafficgen" class seems to be deprecated, I still applied the same adjustments on the uniform traffic.

-------------------------------------------
Demonstrating the bug and the fix
-------------------------------------------

This bug is easy to reproduce with a 4x4 mesh topology, for example with the following config file:
[mesh_bug_reproduce.py](https://github.com/user-attachments/files/22215275/mesh_bug_reproduce.py)

Before this fix, the result shows that the last Endpoint receives no packets:
image

After this fix:
image

### DIFF
--- a/src/sst/elements/merlin/tests/refFiles/test_merlin_torus_5_trafficgen.out
+++ b/src/sst/elements/merlin/tests/refFiles/test_merlin_torus_5_trafficgen.out
@@ -1,1 +1,1 @@
-Simulation is complete, simulated time: 31 ns
+Simulation is complete, simulated time: 38 ns
--- a/src/sst/elements/merlin/tests/torus_5_trafficgen.py
+++ b/src/sst/elements/merlin/tests/torus_5_trafficgen.py
@@ -24,7 +24,7 @@ sst.merlin._params["input_latency"] = "0.0ns"
 sst.merlin._params["output_latency"] = "0.0ns"
 sst.merlin._params["input_buf_size"] = "16.0KB"
 sst.merlin._params["output_buf_size"] = "16.0KB"
-sst.merlin._params["link_lat"] = "5000ns"
+sst.merlin._params["link_lat"] = "1ns"
 
 merlintorusparams = {}
 merlintorusparams["num_dims"]=1


ziyuezzy:
Thanks for double-checking! 
I have removed the changes in autogen.sh, and targeted at branch 'devel' instead.

feldergast:
Looks like you'll need to update this reference file to pass the pre-merge testing:

src/sst/elements/merlin/tests/refFiles/test_merlin_torus_5_trafficgen.out

ziyuezzy:
I updated the test files for test_merlin_torus_5_trafficgen.
However, in the original test config, link latency was set to 5000ns, while the simulation ends in 38ns. As a result, no packet has been received during the simulation (observing from the statistics csv output file).
So I also adjusted the link latency in this test to 1ns, in order to make the result of this test more sensible/meaningful...

feldergast:
> I updated the test files for test_merlin_torus_5_trafficgen. However, in the original test config, link latency was set to 5000ns, while the simulation ends in 38ns. As a result, no packet has been received during the simulation (observing from the statistics csv output file). So I also adjusted the link latency in this test to 1ns, in order to make the result of this test more sensible/meaningful...

Great thanks!

feldergast (approved):

PR #2573, opened 2025-09-10
Vanadis syscalls and cleanup

gvoskuilen:
Adds clone3 and tgkill syscall support. With this, pthreads works under the Clang/rv-gnu-toolchain compiler setup. Vanadis also exits cleanly when the simulated application encounters an error causing a tgkill. 

Also:
* Adds error checking on subcomponent loads in vdecoder.h
* Standardizes var names in vcpuos.h and virtMemMap.h/cc towards the "\_" separation/postfix convention vs camel case and "m_" prefix (all are used throughout vanadis right now and that file used a mix of all conventions).
* Removes unused "cloneReqs" data structure from vanadis.h/cc
* Fixes double loop over hw_threads when using resetRegisterUseTemps function
* Eliminates double decoder loop through hw_threads in favor of a single loop in vanadis rather than the per thread decoders. Adds round-robin mechanism for starvation.
* Add a pendingLoad() function to limit how many times the decoders attempt to re-issue loads for the same instruction.
* Add output pointer to decoder class instead of constantly passing it through function calls
* Add utils.h which currently only has the definition of PRI_ADDR that was previously duplicated in many files. Still needs clean up to eliminate the def from some files.
* Add missing includes to freeList.h
* Changes name of "checkpoint" to "snapshot" in some files (will change the rest later) to avoid confusion - this capability is not SST's checkpoint, it's an application snapshot that can be reloaded.
* Fix unmap so that it handles unmapping part of a region
* Fix mprotect so that it doesn't break linking to program data when part of a program data region is mprotect'd
* Split virtMemMap into header/impl for code organization
* Add output object as a member of VanadisInstructionLoader and replace some asserts with output->fatal()


src/sst/elements/vanadis/inst/vinst.h
```diff
@@ -109,7 +109,7 @@ class VanadisInstruction
             count_isa_fp_reg_in(copy_me.count_isa_fp_reg_in),
             count_isa_fp_reg_out(copy_me.count_isa_fp_reg_out)
         {
-            trapError             = copy_me.trapError;
+            trap_error_           = copy_me.trap_error_;
```
hughes-c, line 112: Don't want to add tails to the rest of the member variables? I think it's fine if you don't since there are lots of other formatting issues with vanadis.

src/sst/elements/vanadis/vanadis.cc
```diff
@@ -1378,7 +1388,7 @@ VANADIS_COMPONENT::tick(SST::Cycle_t cycle)
     if(output_verbosity >= 2)
```
hughes-c, line 1388: The output level shouldn't match this? Is output_verbosity not being passed as part of output (e.g. why this check and then the output)?

src/sst/elements/vanadis/vanadis.cc
```diff
@@ -1417,7 +1427,7 @@ VANADIS_COMPONENT::tick(SST::Cycle_t cycle)
     if(output_verbosity >= 9)
```
hughes-c, line 1427: Same comment about matching the test.

hughes-c:
Looks like we need new reference files due to the pipeline changes.

src/sst/elements/vanadis/vanadis.cc
gvoskuilen: It should match, fixing that. Usually the check is to avoid having to call verbose() in the first place which has non-negligible overhead when you do it every cycle. We should eventually ifdef and/or convert these to output->debug() to eliminate the overhead if you're not printing anything.

src/sst/elements/vanadis/utils.h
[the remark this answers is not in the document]
gvoskuilen: I'm thinking utils.h will be for vanadis-specific stuff and I'll create a syscall_defines.h somewhere in the os/ directory that holds defines that replicate those in linux headers.

src/sst/elements/vanadis/inst/vinst.h
gvoskuilen: I'll update a few more but since this is a base class and protected vars, the changes cascade and a lot of files need updates. Eventually I'll fix the rest.

PR #2592, opened 2025-10-24
Carcosa Element

mgoldstein322:
Adds Carcosa Element to SST Elements
**Features**
1. Generic Fault Injection interface using PortModules
2. Generic Fault interface that plugs into Fault Injectors
3. Out-of-the-box support for stuck bits, corrupt memory regions, random bit flip, and random event drop faults
4. Injectors can be built to perform any number of faults, and any fault can be built to modify MemEvents 

**Known Issues**
- Random bit flip sometimes causes a crash in the simulator when modifying payloads that contain certain errors--unsure if this is solvable without significant retooling in MemEvent or a possible complementary PortModule that listens for specific modifications in events and "fixes" them so that they don't blow up the simulation. Potentially solvable with more complex logic on the bit/byte choosing function. 

**Planned Updates**
1. Add Hali interface for dynamic fault switching and dynamic injector swapping
2. Refine stuck bit injector to be easier to use

### DIFF
--- a/src/sst/elements/carcosa/injectors/dropFlipFaultInjector.cc
+++ b/src/sst/elements/carcosa/injectors/dropFlipFaultInjector.cc
@@ -59,7 +59,7 @@ bool DropFlipFaultInjector::doInjection() {
         this->triggered_injection_[1] = true;
     }
     else {
-#ifdef __SST_DEBUG_OUTPUT__ 
+#ifdef __SST_DEBUG_OUTPUT__
         dbg_->debug(CALL_INFO_LONG, 1, 0, "Flip skipped.\n");
 #endif
         this->triggered_injection_[1] = false;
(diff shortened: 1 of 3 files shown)


bfeinberg:
A few assorted notes:

1) Is there a reason `FaultBase` isn't in the main Carcosa directory like `FaultInjectorBase`? I feel like both bases classes should be there or neither.
2) I don't like the Name `DoubleFaultInjector` it makes me thing this is a fault injector for doubles not two types. I'd call it `DropFlipFaultInjector`
3) I think it might be clearer to have the `FaultInjectorBase`class not have a fault probability parameter then each FI class can do their own thing. Notably stuckAt and corruptMem ignore this param, and the way "Double" behaves to make use of this parameter just makes things more confusing. For Double it is probably eaiser to have two rates, occasionally they'll hit the same value, but thats not a non-physical behavior.
4) Make the default FI rate 0. 0.5 is unrealistically high and there isn't really a sane default so for consistency with the others just go with 0.
5) think we need some additional documentation on the interaction between fault logic and fault injectors. I understand it because we talked about, but just a paragraph explaining this in FaultInjectorBase.h
6) There are some stats hooks in FaultInjectorBase commented out. We should either implement these or remove them.

bfeinberg:
Changes look good to me.

feldergast (approved):
Approving based on @bfeinberg's review.

Issue #2584, opened 2025-10-08
Potential PR: source routing in Merlin?


ziyuezzy:
# This is a issue made for a potential PR:
Dear devs,

I am working on a feature for SST-Merlin. Because it contains a considerable amount of work, I would really appreciate your opnions on whether or not this is a Pull Request that you will consider to merge. Any comments on the current design approach is welcome.

## Objective

The main goal is to support source-routing in SST-merlin. Source routing means the routed path will be determined before a packet enters the first router or switch. 

This will require the network admin to be aware of the network topology, and there should be algorithms (e.g., in the endpoint NICs) to determine the routed paths for the given network topology.

## Design

In additional to the original merlin module, the following additions will be made:

1. Add source-routing support in routers (implemented throught the topology modules).

- For this, I am now adopting an input-from-file method. So the user would need to generate a `.graphml` file (for example, using the networkx library in python) to describe the network topology and input that into the SST config file. 

- A new topology class is now implemented , namely `topo_any`, which will parse the input graph description and build the network accordingly. Note that the main motivation was to support source routing. However, topology-independent Destination-tag routing would also be possible with the `topo_any` topology class. 
    - Source routing: it will require the encapsulated request can be dynamic_cast into a `SRRequest` (will be explained later) that has an embedded routed path. What the router do in this case is basically reading the next-hop router id defined in the embedded routed path and forward.
    - Destination-tag routing: it will require another input file to specify the routing table (for each router, a mapping from the destination EP id to the next-hop router id). Building on this, topology-independent ECMP or UGAL routing algorithms can be implemented. For topology-specific destination-tag routing algorithms, one should still use that specific topology class.

2. Add source-routing support in the endpoints.
- For this, I am defining a new class `endpointNIC` which is a child class of `SST::Interfaces::SimpleNetwork`. It interacts with two sides -- a traffic generator and a linkcontrol module. This `endpointNIC` class will build a framework to enable NIC features at the endpoins, while its child classes will further extend NIC-specific functionalities, as will be explained later.
    - On the endpoint side, 'endpointNIC' talks to a traffic generator which can be any module that generates traffic for merlin, e.g., `offeredload`, `simple_patterns/*`, `test/nic`, `trafficgen`, `emberjob`, etc.
    - On the network side, 'endpointNIC' holds either a LinkControl or a ReorderLinkControl as a subcomponent, delegating all sends and receives to it.
[...]

PR #2598, opened 2025-11-04
Anytopo

ziyuezzy:
Hi Devs,

This is a pull request, previously mentioned in [issue #2584 ](https://github.com/sstsimulator/sst-elements/issues/2584)

There are documentations of this work in 
`"sst-elements-src/src/sst/elements/merlin/interfaces/endpointNIC/documentations/README.md"`, and 
`"sst-elements-src/src/sst/elements/merlin/topology/anytopo_ultility/README.md"`

In a few words, this branch extends the Merlin module in SST-element, such that any network topology can be built with an input networkx graph. Another key feature is to support source routing (the endpoint NIC determines the routed paths). And a few popular HPC network topologies (dragonfly, slimfly, polarfly, jellyfish) are already defined in 
`"sst-elements-src/src/sst/elements/merlin/topology/anytopo_ultility/"` dir, together with some tests in `"sst-elements-src/src/sst/elements/merlin/topology/anytopo_ultility/tests"`. 

If you would like to accept this pull requests, maybe we can move these tests to the main merlin test dir?

It worth mention that the reorderedlinkcontrol has been modified to fit in the new 'ExtendedRequest' framework, this has been included in tests as well, see` EndpointNIC(use_reorderLinkControl=True ...` in the tests.

I believe that this pull request will have the following contribution to Merlin:
- supporting more network topologies (any topology from input graph)
- supporting source routing, with is increasingly interesting in HPC network traffic engineering
- Adds a framework of EndpointNIC that allows different NIC/smartNIC functionalities to be plugged in the packet-processing pipeline. For now source routing is implemented through this framework.

Please let me know if there is any feedback/comments/questions for the code.

Best regards,
Z.


feldergast:
In order to merge the PR, the tests will need to be part of the main merlin testing.  Go ahead and move them there and we can review the code and release the PR for testing.

ziyuezzy:
Before I add new tests, I want to make sure the current tests all pass.
I don't really understand why the 'macos-14 / macos-15' are failing here? 

When I run 'sst-test-elements' locally (ubuntu 22, python3.8), all tests pass (see screenshot below). 
image

For example the test "test_EmberSweep_*" all passed locally, but failed in the github action... 
Does anyone have an idea what is wrong there?

feldergast:
There are some known issues with the GitHub macOS runners and how they interact with SST.  You can safely ignore the results of the GitHub macOS builders for now.  They are experimental and the official testing is done on other machines.

feldergast:
Looks like that those issues were actually fixed a week or so ago.  I dug a little deeper and it looks like you introduced a dependance on a third-party library that isn't installed on those systems (NetworkX).  You'll need to do a check for that and not load it if it's not there (i.e. anytopo won't be available if NetworkX isn't there.  The polarfly and polarstar topos have a similar issue with a different library.  Give me a bit to look through the code and I can send some pointers on the best way to do this.

src/sst/elements/merlin/topology/pymerlin-topo-any.py
```diff
@@ -0,0 +1,298 @@
...
+See pymerlin-interface.py for SourceRoutingPlugin usage and routing table serialization.
+"""
+
+import sst
+from sst.merlin.base import *
+import os
+from collections import defaultdict
+from typing import Callable, Optional
+
+try:
```
feldergast: This looks like the code segment that is failing when networks isn't installed.  It looks like this code is executing when the module is loaded, so the ImportError below ends up failing in the module loading.  So, this causes every simulation that imports sst.merlin to fail, not just those that use topoany.  Probably need to set a variable if the modules are missing and fail when trying to use the topology instead of at module load time.

feldergast:
Looks like polarstar and polarfly depend on the same third-party Python libraries as anytopo, so we already have builders set up with those libraries, so anytopo should test on those and be skipped on all the ones without those libraries.

ziyuezzy:
> Looks like polarstar and polarfly depend on the same third-party Python libraries as anytopo, so we already have builders set up with those libraries, so anytopo should test on those and be skipped on all the ones without those libraries.

Ok that's good news. Thanks for checking. 
Will the github action automatically skip or continue with the tests, depending on whether or not networkx is installed? Or Should I add some indications in the test files?

Another question is that some of my tests require extra python libraries, for example I imported the library "galois" for generating the 'Slimfly; topology. This import is not built with SST, but only is required when running the test.  
Is there any solution there?

feldergast:
You'll need to add an @unittest.skipIf similar to what the polar* topology tests do so that they don't try to execute if the libraries aren't found. Note that the testsuite_Default_merlin.py file also tries to load the required libraries so that they can check if they are in sys.modules.

ziyuezzy:
I have added new tests for Anytopo, and all tests passed. So I think this PR is ready for review.

feldergast:
Sorry, with the SC conference last week, Thanksgiving this week and vacation next week, it will be a bit before I can dive deep into this PR.  I took a quick glance through it and didn't see anything that immediately stood out as an issue.  I'll revisit this the week of December 8th and do a complete review.

feldergast:
Sorry for the delay, things got a bit hectic during December.  I will be looking at this PR this week, though it may take me a bit to get through since I'll have to remind myself of the details in the merlin code.

src/sst/elements/merlin/interfaces/endpointNIC/NICPlugin.h
```diff
@@ -0,0 +1,41 @@
...
+        SST::Interfaces::SimpleNetwork::Request* req, int vn) = 0;
+
+    // Get plugin name for debugging
+    virtual std::string getPluginName() const = 0;
+
+    // Optional: called during init/setup/finish phases
+    virtual void plugin_init(unsigned int phase) {}
+    virtual void plugin_setup() {}
+    virtual void plugin_finish() {}
+};
```
feldergast: missing plugin_complete()

src/sst/elements/merlin/interfaces/endpointNIC/SourceRouting.h
```diff
@@ -0,0 +1,138 @@
...
+    ~WeightedRoundRobinSelection();
+    std::deque<int> selectPath(int dest_router, const std::vector<path_with_weight>& paths) override;
+    void reset() override;
+};
+
+class SourceRoutingPlugin : public NICPlugin
+{
+public:
+    // Static shared data accessible by both SourceRoutingPlugin and topology classes
+    static std::map<int, int> endpoint_to_router_map;  // Shared endpoint-to-router mapping
```
feldergast: Generally, static variables shouldn't be used in SubComponents.  They should use the SharedObjects provided by the core.  Since both of the classes that need access are derived from SubComponent, they will have access to those APIs.

ziyuezzy:
Hi, no worries, we are not in any hurry. 
And thanks for giving feedbacks on the code, I will try to fix the points you raised.

Recently I have been thinking about the design of this branch. 
Of course importing network topology from graph, and making the routers do source-routing, is probably necessary to be implemented in Merlin. 
But this whole concept of 'EndpointNIC with plugins', feels suitable to be an individual sst-element, similar to the 'rdmaNic' element?
Implementing these in Merlin, may benefit its extensibility. For example, I recently implemented another small plugin (not in this branch) that dumps traffic traces from Merlin, it was really clean and efficient. I wonder whether it is even feasible to implement circuit-switching within this framework....
But on the other hand, with a lot of plugins, this framework might also make Merlin complicated/confusing to read for people just started to use this simulator... I don't know if that is a concern or not.

src/sst/elements/merlin/interfaces/endpointNIC/SourceRouting.h
feldergast: The SharedObjects pretty much support any Object that is serializable, which means that a mutex would not be supported.  If you need to have a shared Mutex, then that will still need to be a static variable.  I/O generally doesn't fall under the same restrictions as variables used during the simulation.

src/sst/elements/merlin/interfaces/endpointNIC/SourceRouting.cc
```diff
@@ -0,0 +1,343 @@
...
+
+RandomWeightedSelection::RandomWeightedSelection(ComponentId_t cid) {
+    rng = new SST::RNG::MersenneRNG(cid);
+}
+
+RandomWeightedSelection::~RandomWeightedSelection() {
+    delete rng;
+}
+
+std::deque<int> RandomWeightedSelection::selectPath(int dest_router, const std::vector<path_with_weight>& paths) {
```
feldergast: Do you have to return this by value, causing a copy of the deque?  If the contents of the deque are only read after the function is called, can you return a const reference instead?

src/sst/elements/merlin/interfaces/endpointNIC/SourceRouting.cc
```diff
@@ -0,0 +1,343 @@
...
+    return paths.back().second;
+}
+
+WeightedRoundRobinSelection::WeightedRoundRobinSelection() {
+}
+
+WeightedRoundRobinSelection::~WeightedRoundRobinSelection() {
+}
+
+std::deque<int> WeightedRoundRobinSelection::selectPath(int dest_router, const std::vector<path_with_weight>& paths) {
```
feldergast: const std::deque<int>& instead??

feldergast:
Just an update.  I've worked through about a third of the code changes, but have a deadline this week and then have a week of vacation.  I will continue to review code when I get back.

deanchester:
One thing that would really improve the usability of these new components is filling out the ELI parameter documentation. Currently `SST_ELI_DOCUMENT_PARAMS()` in `anytopo.h` is empty, and `endpointNIC.h` only documents `plugins` despite the components accepting several other parameters.

The ELI documentation is what users see through `sst-info` and is often the first place people look when trying to configure a component — without it, the only way to figure out valid parameters and their expected values is to read the source. Filling these in would make the new topology and NIC framework much more approachable.

For reference, here's what the `anytopo.h` `SST_ELI_DOCUMENT_PARAMS` could look like, following the style used by the existing dragonfly topology:

```cpp
SST_ELI_DOCUMENT_PARAMS(
    {"num_routers",                "Number of routers in the network."},
    {"num_R2R_ports",              "Number of router-to-router ports on this router."},
    {"num_R2N_ports",              "Number of router-to-endpoint (NIC) ports on this router."},
    {"routing_mode",               "Routing mode per VN [source_routing | dest_tag_routing].", "source_routing"},
    {"source_routing_algo",        "Source routing algorithm per VN [weighted | UGAL | UGAL_THRESHOLD].", "weighted"},
    {"vcs_per_vn",                 "Number of virtual channels per virtual network.", ""},
    {"vn_ugal_num_valiant",        "Number of valiant paths to consider for UGAL routing per VN.", "3"},
    {"adaptive_threshold",         "Threshold for UGAL_THRESHOLD adaptive routing decisions.", "2.0"},
    {"connectivity",               "Router connectivity string. Format: \"dest_id,port0,port1;dest_id,port0,...\""},
    {"simple_routing_entry_string","Serialized routing table for untimed packets. Format: \"dest:weight,hop1,hop2;weight,hop3|dest:...\""},
    {"endpoint_to_port_map",       "Map of global endpoint IDs to local router port numbers."},
    {"port_to_endpoint_map",       "Map of local router port numbers to global endpoint IDs."},
    {"verbose_level",              "Verbosity level for debug output.", "0"},
)
```

Similarly, `endpointNIC.h` should document `EP_id`, `networkIF`, and `plugin_names` alongside the existing `plugins` entry.

src/sst/elements/merlin/interfaces/endpointNIC/endpointNIC.h
```diff
@@ -0,0 +1,100 @@
...
+#include <sst/core/link.h>
+#include <sst/core/output.h>
+#include "sst/core/interfaces/simpleNetwork.h"
+#include "NICPlugin.h"
+#include <vector>
+
+namespace SST {
+namespace Merlin {
+
+class endpointNIC : public SST::Interfaces::SimpleNetwork
```
feldergast: Generally the class name would capitalized camel case, i.e. EndpointNIC

src/sst/elements/merlin/interfaces/endpointNIC/endpointNIC.cc
```diff
@@ -0,0 +1,232 @@
...
+    // See if we need to set up a nid map
+    bool found = false;
+    EP_id = params.find<int>("EP_id",-1,found);
+    if ( !found ) {
+        out.fatal(CALL_INFO, -1, "endpointNIC requires 'EP_id' parameter to be set\n");
+    }
+
+    loadPlugins(params);
+
+    std::string networkIF = params.find<std::string>("networkIF");
```
feldergast: Looks like this is not used other then in the commented out code below.

src/sst/elements/merlin/interfaces/endpointNIC/endpointNIC.cc
```diff
@@ -0,0 +1,232 @@
...
+endpointNIC::endpointNIC(ComponentId_t cid, Params& params, int vns) :
+    SST::Interfaces::SimpleNetwork(cid),
+    link_control(nullptr),
+    vns(vns)
+{
+
+    out.init(getName() + ": ", 0, 0, Output::STDOUT);
+    // See if we need to set up a nid map
+    bool found = false;
+    EP_id = params.find<int>("EP_id",-1,found);
```
feldergast, line 32: I think for the anytopo topology, getting the endpoint ID from params makes sense, but if you're using this with any of the other topologies as a way to do a NIC with extra plugins, then the EndpointNIC will need to get the ID from the topology using LinkControl::getEndpointID during init().  This may necessitate a change in some of the plug-in APIs so that the endpoint ID can be assigned later.

I don't think this needs to be done before an initial merge, but some thought should be given to how this could be done to make this new infrastructure more generally useful.

src/sst/elements/merlin/interfaces/endpointNIC/endpointNIC.cc
```diff
@@ -0,0 +1,232 @@
...
+                return nullptr;
+            }
+        }
+    }
+
+    return current_req;
+}
+
+} // namespace Merlin
+} // namespace SST
```
feldergast: Needs new line at end of file

src/sst/elements/merlin/interfaces/endpointNIC/endpointNIC.h
```diff
@@ -0,0 +1,100 @@
...
+    void loadPlugins(Params& params);
+
+    // Process request through plugin pipeline
+    Request* processThroughPipeline(Request* req, int vn, bool outgoing);
+};
+
+} // namespace Merlin
+} // namespace SST
+
+#endif
```
feldergast: New line

src/sst/elements/merlin/interfaces/endpointNIC/sourceRouting.cc
```diff
@@ -0,0 +1,356 @@
...
+{
+}
+
+SST::Interfaces::SimpleNetwork::Request* SourceRoutingPlugin::processOutgoing(
+    SST::Interfaces::SimpleNetwork::Request* req, int vn)
+{
+    if (!req) return nullptr;
+
+    // Convert to ExtendedRequest if not already
+    ExtendedRequest* ext_req = dynamic_cast<ExtendedRequest*>(req);
```
feldergast: Dynamic casts can have a noticeable performance impact.  I'm not sure there's a way to avoid this given the core APIs being used, but we should think of enhancements to the SimpleNetwork interface that could remove the need for a dynamic_cast().

src/sst/elements/merlin/interfaces/endpointNIC/SourceRouting.cc
```diff
@@ -0,0 +1,356 @@
...
+        delete req;
+    }
+
+    // Check if path metadata already exists
+    SourceRoutingMetadata sr_meta;
+    if (!ext_req->getMetadata("SourceRouting", sr_meta) || sr_meta.path.empty()) {
+        // Lookup path in routing table and set it as metadata
+        std::deque<int> path = selectPath(lookupRtrForEndpoint(ext_req->dest));
+
+        // // use verbose to print out this routing decision:
```
feldergast: If this code isn't needed, it should be removed.  If you want to keep it around, it should use the Output object instead cout.

src/sst/elements/merlin/interfaces/endpointNIC/SourceRouting.h
```diff
@@ -0,0 +1,137 @@
...
+    std::deque<int> selectPath(int dest_router);
+    inline size_t lookupRtrForEndpoint(SST::Interfaces::SimpleNetwork::nid_t endpoint) {
+        return endpoint_to_router_shared[endpoint];
+    }
+};
+
+} // namespace Merlin
+} // namespace SST
+
+#endif
```
feldergast: New line

src/sst/elements/merlin/interfaces/reorderLinkControl.cc
```diff
@@ -121,27 +122,29 @@ void ReorderLinkControl::finish(void)
 bool ReorderLinkControl::send(SimpleNetwork::Request* req, int vn) {
     if ( vn >= vns ) return false;
     if ( !link_control->spaceToSend(vn, req->size_in_bits) ) return false;
-    ReorderRequest* my_req = new ReorderRequest(req);
-    delete req;
 
-    // Need to put in the sequence number
+    // Convert to ExtendedRequest if not already
+    Merlin::ExtendedRequest* ext_req = dynamic_cast<Merlin::ExtendedRequest*>(req);
```
feldergast: I'm worried about potentially having to have multiple dynamic_casts() for every request that is sent.  If you were to have en EndpointNIC that needed a ReorderLinkControl that also has a LinkControl, then you end up multiple calls to dynamic_cast().  When I was architecting the event hierarchy for merlin, there was a noticeable performance difference between dynamic casting to determine event type and putting an enum in the base class to query the type.  This type of structure is little more difficult given the fact that the base class is a core class, but I'm not comfortable with the dynamic casts that will potentially slow down simulations that don't even use the plug-in capable EndpointNIC.

src/sst/elements/merlin/interfaces/reorderLinkControl.h
```diff
@@ -35,63 +35,77 @@ class Component;
...
-        {
-            givePayload(req->takePayload());
-            trace = req->getTraceType();
-            traceID = req->getTraceID();
-        }
+// public:
+//     uint32_t seq;
 
-    ~ReorderRequest() {}
+//     ReorderRequest() :
```
feldergast: Due to all the additional dynamic casts, I'm not convinced this is a good path to take.

feldergast (changes requested):
I've looked through all the code.  Generally it looks good, with just a few minor code cleanup suggestions. However, there is one big issue that needs to have some thought put into it, which is the fact that the modified code structure has added a series of dynamic_casts() to the heavily used code paths.  We have seen noticeable performance issues when dynamic_cast() is used often.  We need to give some thought to limiting the use of dynamic casts, which may require some changes to the SimpleNetwork API in the core. I will need to give this some more thought after the SST 16 release goes in the next week or so.  I had hoped to get this into the release, but I don't know that we can resolve this issue before the release.

ziyuezzy:
Hi, thanks for your feedbacks, I have incorporated them in the code.

Regarding the dynamic_casts, I am optimizing with the following modifications: 

- In ReorderLinkControl: 
Unconditioanlly instantiate, avoiding dynamic cast:
```c++

    Merlin::ExtendedRequest* ext_req = new Merlin::ExtendedRequest(req);
    delete req;

```
- In endpointNIC:
Use one dynamic cast before the loop of plugin pipeline
- In sourceRouting and Anytopo:
Use static cast instead of dynamic cast
```c++

    assert(dynamic_cast<ExtendedRequest*>(req) != nullptr);
    ExtendedRequest* ext_req = static_cast<ExtendedRequest*>(req);

```
So the dynamic_cast will only run with debug flags, while only static_cast will be executed with -O0 or -O1 compile flag... Do you agree with this approach?

Btw, I don't know why the macos ci is blocking the all-tests now... But I ran locally 'sst-test-elements' there were no failure.

feldergast:
Sorry, the notification for this got lost in a flurry of other emails.  I should be able to look through this again next week.  However, it looks like there are now merge conflicts, so you'll need to rebase your changes on the current devel branch.  I suspect the addition of checkpointing to merlin is the culprit, but haven't looked specifically at what the conflicts are.

ziyuezzy:
I have updated the branch, but the CI action is not triggered.. I don't know why.

ziyuezzy:
I have built and ran tests locally, all passed

feldergast:
Looks like at least one header file (ExtendedRequest.h) is missing from Makefile.am so the make dist test is failing.  Please make sure any files that were added are in Makefile.am.  If you have any questions about where specific files belong in the file, let me know.

feldergast (approved):

PR #2603, opened 2025-11-26
Vanadis: Enable true Out-of-Order execution for RoCC instructions

### LINKED ISSUE #2602: Vanadis - Enable true Out-of-Order execution for RoCC instructions by fixing register renaming and dispatch logic
**Description**
Hi, I am Yifan Zhang from the ChatDV group at Southeast University. I am currently working on integrating a custom CIM accelerator with Vanadis using the RoCC interface for our research.

During performance modeling, I identified two critical limitations in the current `VanadisRoCCInterface` integration that prevent true Out-of-Order (OoO) execution, effectively forcing RoCC instructions to execute sequentially or stall the pipeline unnecessarily.

**Current Limitations**

1.  **Early Source Operand Evaluation (RAW Dependency Issue):**
    Currently, `assignRegistersToInstruction` reads the *values* of source registers (`rs1`, `rs2`) immediately during the Issue/Rename stage and pushes them to the RoCC interface.
    *   *Problem:* In a realistic OoO core, the values might not be ready yet (e.g., pending a previous ALU or Load instruction). Reading them at the Issue stage forces the pipeline to wait until the producer instruction retires or forces the RoCC instruction to execute with stale data if not handled correctly. It bypasses the standard reservation station/waiting logic.
    
    Here is the problematic logic in assignRegistersToInstruction. It reads register values immediately, bypassing the reservation station/dependency check mechanism:
```cpp
// In VANADIS_COMPONENT::assignRegistersToInstruction

if (ins->getInstFuncType() >= INST_ROCC0 && ins->getInstFuncType() <= INST_ROCC3) {
    // ...
    if (!roccs_[rocc_index]->RoCCFull()) {
        VanadisRegisterFile* regFile = register_files[ins->getHWThread()];
        
        // PROBLEM: Values are read immediately at the Issue/Rename stage!
        // These physical registers might be waiting for a producer (RAW dependency).
        uint64_t rs1_val = regFile->getIntReg<int64_t>(ins->getPhysIntRegIn(0));
        uint64_t rs2_val = regFile->getIntReg<int64_t>(ins->getPhysIntRegIn(1));

        // ...
        // Instruction is pushed to accelerator before operands are actually ready
        roccs_[rocc_index]->push(new RoCCCommand(rocc_inst, rs1_val, rs2_val));
    }
}
```

2.  **Lack of Destination Register Renaming (WAW/WAR Dependency Issue):**
    RoCC instructions currently write back results directly to the architectural register (`rd`) index upon completion, without allocating a new physical register from the free list during the Rename stage.
    *   *Problem:* This re-introduces Write-After-Write (WAW) and Write-After-Read (WAR) hazards, causing unnecessary serialization and preventing the CPU from overlapping independent instructions with long-latency accelerator tasks.

    And here is the write-back logic in performExecute. It writes directly to resp->rd (architectural index) instead of a mapped physical register:
```cpp
// In VANADIS_COMPONENT::performExecute

// Tick the RoCC Interfaces
for (int i = 0; i < roccs_.size(); i++) {
    RoCCResponse* resp;
    if (!(roccs_[i]->isBusy()) && (resp = roccs_[i]->respond())) {
        
        VanadisInstruction* ins = rocc_queues_[i].front();
        
        // PROBLEM: Writing to architectural register 'rd' directly.
        // This ignores register renaming and causes WAW/WAR hazards.
        register_files[ins->getHWThread()]->setIntReg<uint64_t>(resp->rd, resp->rd_val);
        
        ins->markExecuted();
        rocc_queues_[i].pop_front();
    } 
    // ...
}
```

**Proposed Solution**

I have prototyped and verified a fix locally that aligns RoCC instruction handling with standard Vanadis functional units. The proposed changes are:

1.  **Implement Register Renaming for RoCC:**
    Modify `assignRegistersToInstruction` to treat RoCC instructions like standard integer instructions. Allocate a physical register for `rd` from the `int_register_stack` and update the ISA table. This resolves false dependencies.

2.  **Delayed Dispatch Mechanism:**
    *   Remove the immediate `push` to the RoCC interface from the Issue stage.
    *   Introduce a `rocc_wait_queue` (similar to a reservation station) to hold issued RoCC instructions.

3.  **Operand Readiness Check in Execute Stage:**
    Modify `performExecute` to check the `rocc_wait_queue`. An instruction should only be dispatched (pushed) to the RoCC interface when:
    *   The RoCC command queue is not full.
    *   **Crucially:** All source physical registers are ready (checking `pendingIntWrites`).
    *   Once dispatched, the register values are read from the register file at this correct moment.

4.  **Correct Write-back:**
    Update the response handling logic to write the result into the allocated *physical* register instead of the architectural register.

**Impact**
These changes allow the Vanadis CPU to continue issuing and executing independent instructions while a RoCC instruction is waiting for operands or executing a long-latency task. This is essential for accurate performance modeling of heterogeneous systems (e.g., CPU + Accelerator).

**Status**
I have implemented these changes in a local fork and verified them with a custom CIM component. I plan to clean up the code (separating it from my specific component logic) and submit a Pull Request in the near future. I wanted to open this issue to track the feature and gather any feedback on the design approach.

yuebanabn:
Fixes #2602 

**Summary**
This PR overhauls the `VanadisRoCCInterface` integration within the processor pipeline to support true Out-of-Order (OoO) execution. Previously, RoCC instructions were handled sequentially with immediate operand reading at the Issue stage and direct architectural register write-back, which artificially serialized execution and introduced RAW/WAW/WAR hazards.

**Technical Details**

This PR implements the following changes to align RoCC instruction handling with standard Vanadis functional units:

1.  **Register Renaming for RoCC (`assignRegistersToInstruction`)**:
    *   Removed the special handling block that bypassed register allocation.
    *   RoCC instructions now go through the standard output register allocation path, acquiring physical registers for `rd`. This resolves WAW and WAR hazards.

2.  **Delayed Dispatch Mechanism (`allocateFunctionalUnit`)**:
    *   Removed the immediate `push` to the RoCC interface during the Issue stage.
    *   Introduced `rocc_wait_queues_` to hold issued RoCC instructions that are waiting for source operands (acting as a reservation station).

3.  **Operand Readiness Check (`performExecute`)**:
    *   Added logic to the start of the Execute stage to scan `rocc_wait_queues_`.
    *   Instructions are only dispatched (pushed) to the accelerator when:
        *   The hardware RoCC queue is not full.
        *   All source physical registers are ready (checking `pendingIntWrites`).
    *   Source register values are now read from the register file at the moment of dispatch, ensuring correct data is sent.

4.  **Correct Physical Register Write-back**:
    *   Updated the response handling logic in `performExecute`.
    *   Results are now written to the allocated **physical register** index (retrieved from the instruction object) instead of the architectural register index.

**Impact**
These changes allow the Vanadis CPU to continue issuing independent instructions while RoCC instructions are waiting for operands or executing long-latency tasks. This significantly improves the accuracy of performance modeling for heterogeneous systems.

**Testing**
*   Verified with a custom CIM (Compute-in-Memory) RoCC component.
*   Confirmed that RAW dependencies are correctly respected (instruction waits for operands).
*   Confirmed that independent instructions can execute out-of-order relative to RoCC operations.

### DIFF
--- a/src/sst/elements/vanadis/vanadis.cc
+++ b/src/sst/elements/vanadis/vanadis.cc
@@ -277,6 +277,7 @@ VANADIS_COMPONENT::VANADIS_COMPONENT(SST::ComponentId_t id, SST::Params& params)
             if (lists->isPopulated(i)) {
                 roccs_.push_back(lists->create<SST::Vanadis::VanadisRoCCInterface>(i, ComponentInfo::SHARE_NONE));
                 rocc_queues_.push_back(std::deque<VanadisInstruction*>());
+                rocc_wait_queues_.push_back(std::deque<VanadisInstruction*>());
                 output->verbose(CALL_INFO, 1, 0, "Successfully loaded RoCC Interface");
             }
         }
@@ -734,6 +735,50 @@ VANADIS_COMPONENT::performIssue(const uint64_t cycle, int hwThr, uint32_t& rob_s
 void
 VANADIS_COMPONENT::performExecute(const uint64_t cycle)
 {
+    // Process RoCC instruction dispatching
+    for (int i = 0; i < roccs_.size(); ++i) {
+        if (rocc_wait_queues_[i].empty()) {
+            continue;
+        }
+
+        if (roccs_[i]->RoCCFull()) {
+            continue;
+        }
+         
+        VanadisInstruction* ins_to_dispatch = rocc_wait_queues_[i].front();
+        const uint32_t hw_thr = ins_to_dispatch->getHWThread();
+
+        // --- Ready to dispatch ---
+
+        // 1. pop instruction from wait queue
+        rocc_wait_queues_[i].pop_front();
+
+        // 2. read value from source registers
+        VanadisRegisterFile* regFile = register_files[hw_thr];
+        uint64_t rs1_val = 0;
+        uint64_t rs2_val = 0;
+
+        if (ins_to_dispatch->countISAIntRegIn() > 0) {
+            rs1_val = regFile->getIntReg<uint64_t>(ins_to_dispatch->getPhysIntRegIn(0));
+        }
+        if (ins_to_dispatch->countISAIntRegIn() > 1) {
+            rs2_val = regFile->getIntReg<uint64_t>(ins_to_dispatch->getPhysIntRegIn(1));
+        }
+
+        // 3. create RoCC instruction and push to RoCC interface 
+        VanadisRoCCInstruction* vrocc_inst = static_cast<VanadisRoCCInstruction*>(ins_to_dispatch);
+        RoCCInstruction* rocc_inst = new RoCCInstruction(
+            vrocc_inst->func7, vrocc_inst->rd, vrocc_inst->xs1, vrocc_inst->xs2, vrocc_inst->xd
+        );
+        roccs_[i]->push(new RoCCCommand(rocc_inst, rs1_val, rs2_val));
+
+        // 4. push instruction to trace queue
+        rocc_queues_[i].push_back(ins_to_dispatch);
+
+        output->verbose(CALL_INFO, 8, 0, "[RoCC Dispatch] Instruction 0x%" PRI_ADDR " dispatched to RoCC-%d.\n",
+            ins_to_dispatch->getInstructionAddress(), i);
+    }
+
 
     #ifdef VANADIS_BUILD_DEBUG
     const uint32_t verbose_level = output->getVerboseLevel();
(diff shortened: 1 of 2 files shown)


bfeinberg:
This may exacerbate a long-standing issue with the Vanadis-RoCC interface, potential speculative execution of RoCC instructions. If you look at the BOOM core documentation it states that RoCC instructions should now issue until they become non-speculative (https://docs.boom-core.org/en/latest/sections/reorder-buffer.html#point-of-no-return-pnr, https://docs.boom-core.org/en/latest/sections/execution-stages.html#the-rocket-custom-co-processor-interface-rocc) ensuring no problems from instructions with side effects.

This has been something I've been meaning to fix for a while but it hasn't been a huge issue because as you point out the RoCC instructions pretty aggressively stall the pipeline. Could you incorporate that non-speculative dispatch into these changes?

yuebanabn:
Thanks for the excellent feedback and pointing out this critical issue. I agree completely.
I'll work on implementing the non-speculative dispatch from the commit stage to correctly handle instructions with side effects.
I'll start by creating a test case to reproduce the bug. I'll keep you updated on my progress.

PR #2604, opened 2025-12-02
MMU: Fix hardcoded line size and general cleanup

### LINKED ISSUE #2567: MMU: Enable custom cache line sizes
https://github.com/sstsimulator/sst-elements/blob/f1fe079ed0aa7269bcff6b4cc06944dcfc1054c8/src/sst/elements/mmu/tlbWrapper.cc#L106

The base address translation is hardcoded to a step size of 64. When the cache uses a line size other than 64, MMU wrapper might cause accesses that straddle two cachelines. 

If the step size can be initialized from a parameter, the MMU configuration can be tuned to match the cache interface's settings.

gvoskuilen:
Fixes #2567 
Some formatting updates to match vanadis
Change "checkpoint" to "snapshot" to differentiate against SST checkpoint
ifdef debug output for faster simulation

### DIFF
--- a/src/sst/elements/mmu/simpleMMU.cc
+++ b/src/sst/elements/mmu/simpleMMU.cc
@@ -101,7 +101,7 @@ void SimpleMMU::map( unsigned pid, uint32_t vpn, uint32_t ppn, int pageSize, uin
 }
 
 void SimpleMMU::map( unsigned pid, uint32_t vpn, std::vector<uint32_t>& ppns, int pageSize, uint64_t flags ) {
-    
+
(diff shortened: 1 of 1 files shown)


src/sst/elements/mmu/simpleMMU.cc
```diff
@@ -129,8 +140,10 @@ void SimpleMMU::dup( unsigned fromPid, unsigned toPid ) {
 }
 
 void SimpleMMU::flushTlb( unsigned core, unsigned hwThread ) {
+#ifdef __SST_DEBUG_OUTPUT__
     m_dbg.debug(CALL_INFO_LONG,1,0,"core=%d hwThread=%d\n",core,hwThread);
 //    sendEvent( getLink(core,"itlb"), new TlbFlushEvent( hwThread ) );
```
hughes-c: Is there a reason that you left this?

gvoskuilen: Nope, I'll clean that up at some point...

PR #2605, opened 2025-12-02
Compile a debug and regular version of vanadis libraries.

gvoskuilen:
Regular version has limited output and is faster. Debug version is what we have been using thus far. This changes the name of the Vanadis core to "vanadis.core" for regular and "vanadisdbg.core" for the debug version. However, "vanadis.dbg_VandisCPU" still works (map to vanadis.core though) so existing scripts will continue to work. Other elements have the same names as before.

All elements in vanadis have both a debug and regular version that can be swapped between by using "vanadisdbg" as the library instead of "vanadis".

For size reasons, we may eventually want to make the default to only compile the regular version of the library.

Code contained in an `#ifdef VANADIS_BUILD_DEBUG" is only compiled in the debug version of the library.

Also fixes a typo: "all processes have exited" instead of "all process have exited".

### DIFF
--- a/src/sst/elements/vanadis/Makefile.am
+++ b/src/sst/elements/vanadis/Makefile.am
@@ -4,11 +4,10 @@
 
 AM_CPPFLAGS += \
 	$(MPI_CPPFLAGS) \
-	-I$(top_srcdir)/src \
-	-DVANADIS_BUILD_DEBUG
+	-I$(top_srcdir)/src
 
 compdir = $(pkglibdir)
-comp_LTLIBRARIES = libvanadis.la
+comp_LTLIBRARIES = libvanadis.la libvanadisdbg.la
 
 VANADIS_SRC_FILES = \
 utils.h \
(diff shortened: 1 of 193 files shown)


src/sst/elements/vanadis/Makefile.am
```diff
@@ -602,13 +601,15 @@ libvanadis_la_SOURCES = \
...
+libvanadisdbg_la_SOURCES = \
+	vanadis.cc \
+	$(VANADIS_SRC_FILES)
 
 libvanadis_la_LDFLAGS = -module -avoid-version
 
-#libvanadisdbg_la_LDFLAGS = -module -avoid-version
+libvanadisdbg_la_LDFLAGS = -module -avoid-version
+
+libvanadisdbg_la_CPPFLAGS = -DVANADIS_BUILD_DEBUG $(AM_CPPFLAGS)
```
hughes-c: Are these flags correct? I think you need to replicate them for both libvanadis_la and libvanadisdbg_la

src/sst/elements/vanadis/os/include/virtMemMap.cc
```diff
@@ -307,9 +307,18 @@ uint64_t VirtMemMap::addRegion(
...
+            assert(0);
+        }
     } else {
-        assert( start = free_list_->alloc( length ) );
+        start = free_list_->alloc( length );
+        if (! start ) {
+            printf("Error: %s, unable to allocate a block of size=%zu\n", name.c_str(), length);
+            std::fflush(stdout);
+            assert(start);
+        }
```
hughes-c: These shouldn't use the sst i/o interface and debug capture rather than assert?

hughes-c (changes requested):

src/sst/elements/vanadis/Makefile.am
hughes-c: @gvoskuilen I think you're right and aree ok with the implicit AM_CPPFLAGS for libvanadis.

hughes-c (approved):

PR #2609, opened 2025-12-11
Fixes for routing untimed broadcasts in the merlin dragonfly topology

rdfriese:
Address two issues in how untimed broadcasts are routed in the dragonfly topology. Before these changes, some hosts would never receive the message while other hosts would receive duplicates of the same message.

1. fixes a off by one error that prevented to last group from receiving messages originating in any other group.
2. fixes resending a the message back to the originating group.

The result is that now the broadcast message arrives to each host exactly once.


rdfriese:
see #2610 for changes on top of up-to-date `devel` branch

PR #2613, opened 2025-12-15
Mercury: remove event library

jmlapre:
Replace all calls to EventLibrary with calls to Library.
Migrate some of the bits over to make everything work.


nab880 (approved):
Looks good, tests that use SumiServer should ensure changes are working correctly.

PR #2616, opened 2025-12-18
added a hybrid-backing mode to membackend backing

nab880:
This adds a hybrid backing the the MemHierarchy memory backing. The hybrid backing contains a malloc backing and a list of mmap backings. Upon receiving a request it first checks if the memory address is within address ranges of the mmap backings, otherwise it accesses using the malloc backing. 

### DIFF
--- a/src/sst/elements/memHierarchy/membackend/backing.h
+++ b/src/sst/elements/memHierarchy/membackend/backing.h
@@ -478,7 +478,7 @@ private:
             Backing* backing = item->myBacking;
             Addr startAddr = item->startAddr;
             Addr endAddr   = item->endAddr;
-            if ( startAddr < addr && addr < endAddr ) {
+            if ( startAddr <= addr && addr < endAddr ) {
                 return backing;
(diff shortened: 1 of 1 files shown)


src/sst/elements/memHierarchy/membackend/backing.h
```diff
@@ -470,6 +470,68 @@ class BackingMalloc : public Backing {
...
+        Backing* backing = getBacking(addr);
+        if ( backing ) {
+            backing->set(addr, size, data);
+        }
+    }
+
+    uint8_t get( Addr addr ) override {
+        Backing* backing = getBacking(addr);
+        if ( backing ) {
+            return backing->get(addr);
```
gvoskuilen: Same as above

feldergast:
Added AT:WIP label to keep our inboxes from filling up over the break since it's unlikely we'll be able to get a review until after the New Year.

gvoskuilen (approved):

PR #2622, opened 2026-01-16
Added the MemEvent Id to the CacheListenerNotification in the CacheListener from MemHierarchy

EweLo:
This PR extends the `CacheListenerNotification` interface to include the ID of the `MemEvent` that triggered the listener callback, enabling custom CacheListener implementations to trace the corresponding MemEvents.

I implemented this because I'm working on a complex tracer in SST that tracks memory access patterns and I therefore need to be able to correlate CacheListener notifications with specific memory events.

- Added `eventId` field and `getEventID()` getter to `CacheListenerNotification`
- Updated all `notifyAccess()` calls across MemHierarchy components to pass the event ID


EweLo:
@gvoskuilen would you be willing to review this? :)

src/sst/elements/memHierarchy/cacheListener.h
```diff
@@ -43,9 +43,10 @@ class CacheListenerNotification {
     CacheListenerNotification(const Addr tAddr, const Addr pAddr, const Addr vAddr,
                               const Addr iPtr, const uint32_t reqSize,
                               NotifyAccessType accessT,
-                              NotifyResultType resultT) :
+                              NotifyResultType resultT,
+                              const MemEventBase::id_type evId) :
```
gvoskuilen: Can you also ensure there is still a constructor that doesn't require the `id_type` field/give id a default value? Other non-memHierarchy libraries use this interface and will break otherwise.

Also change `MemEventBase::id_type` to `SST::Event::id_type`

gvoskuilen (commented):
This change is sounds OK in general but I'd ask for two changes: (1) use `SST::Event::id_type` and `SST::Event::NO_ID` as the type is defined in that class, not memHierarchy and (2) ensure that `event_id` is not a required field since this interface is used by non-memHierarchy libraries and requiring it will break them.

ivlub:
Hi @EweLo , thanks for this PR! I actually really need this feature for my thesis at CAPS now.
Are you still planning to update this? If you are busy, would you mind if I took your commits, applied the requested fixes, and opened a new PR crediting you? Please let me know:)

EweLo:
@gvoskuilen I applied your suggested changes. Would you mind having a look at this again? :)

PR #2632, opened 2026-02-12
Carcosa Interface Components

nab880:
This PR adds the Carcosa interface layer and supporting components so Hali can sit between CPUs/sensors and the memory hierarchy, coordinate fault injection via a shared registry, and support both sensor/ring workloads and Vanadis MMIO coordination (e.g. ping-pong). It also adds the MemHierarchy PortModule faultInjectorMemH, the PMDataRegistry for manager–injector communication, and tests that verify Hali, the manager logic, and fault injection behavior.

Hali – Main interface component. 
Sits on the memHierarchy (highlink = CPU side, lowlink = cache/memory side). Forwards MemEvents 
Uses a FaultInjManager to attach data that PortModules can read to modify behavior.
Example of this working exists in the added faultInjector faultInjectorMemH. Ideally this would be abstracted out into the general "FaultInjector API" but I leave that to future work. 
Working Features:
MMIO mode when control_addr_base/control_addr_size are set: intercepts loads/stores to a control region for Vanadis coordination (command/status registers), with optional ring-based “done” sync (see ping-ping example in Carcosa tests).

FaultInjManagerAPI / FaultInjManager – Subcomponent used by Hali. Queues highlink/lowlink PortModule requests. Uses event ID to coordinate behavior through PMDataRegistry. 

PMDataRegistry – Shared state between manager and injectors.

CarcosaMemCtrl – MemHierarchy-style memory controller with optional iflLinks_N. Used for backing manipulation. 

hyades.h – Single-header API for Vanadis to talk to Hali’s MMIO control region. 
Applications that use hyades should have a jump table Upon accessing the MMIO control region the Hali interface intercepts the memory access and returns a jump table index.  See pingpong.c in carcosa/tests for example. This will likely need to be expanded for more complex use cases. 

testCarcosaPingPong.py – Two Vanadis cores, one VanadisNodeOS, two processes (same binary, roles 0 and 1). Puts Hali in each core’s data path (CPU → Hali → dTLB → L1D), MMIO params

### DIFF
--- a/src/sst/elements/carcosa/hyades.h
+++ b/src/sst/elements/carcosa/hyades.h
@@ -1,0 +1,158 @@
+/*
+ * Hyades - Vanadis control abstraction
+ *
+ * Single header that encapsulates the MMIO coordination protocol used when
+ * running under SST/Vanadis with a Hali in the data path. Hali intercepts
+ * accesses to a small MMIO region and coordinates multiple cores (e.g. ping-pong).
+ *
+ * -----------------------------------------------------------------------------
+ * BUILD-TIME PARAMETERS (define before #include "hyades.h" if not using default)
+ * -----------------------------------------------------------------------------
+ *
+ *   HYADES_MMIO_BASE
+ *     Optional. Base virtual address of the MMIO control region. Must match
+ *     the Hali parameter "control_addr_base" in the SST config.
+ *     Default: 0xBEEF0000
+ *     Example: #define HYADES_MMIO_BASE 0xBEEF0000
+ *
+ * -----------------------------------------------------------------------------
+ * PROTOCOL (must match Hali MMIO handling)
+ * -----------------------------------------------------------------------------
+ *
+ *   - Base + 0x0 (command register, read):
+ *       The CPU reads the next action index. Value < 0 means "exit" (end the
+ *       run loop). The read may block until Hali has a command ready.
+ *
+ *   - Base + 0x4 (status register, write):
+ *       The CPU writes the completed action index to signal done. Hali
+ *       coordinates with the partner core, then advances and may unblock
+ *       the next command read.
+ *
+ * -----------------------------------------------------------------------------
+ * RUNTIME API
+ * -----------------------------------------------------------------------------
+ *
+ *   hyades_command_read(void)
+ *     Returns the next command index (>= 0) or exit sentinel (< 0).
+ *     Parameters: none.
+ *
+ *   hyades_status_write(int idx)
+ *     Signals completion of the action for index idx.
+ *     Parameters: idx - the command index that was just executed.
+ *
+ *   hyades_run(hyades_handler_t *handlers, int n_handlers)
+ *     Runs the control loop: read command -> call handler -> write status,
+ *     until the command is exit. Required for the standard use pattern.
+ *     Parameters:
+ *       handlers   - array of function pointers (jump table); may be NULL for
+ *                    unused indices.
+ *       n_handlers - number of valid entries in handlers (e.g. 2 for ping/pong).
+ *
+ *   hyades_role_from_argv(int argc, char *argv[])
+ *     Optional helper. Parses role from argv[1] (e.g. "0" or "1"). Used to
+ *     select which action map this process uses (e.g. role 0: [ping, pong],
+ *     role 1: [pong, ping]).
+ *     Parameters: argc, argv - standard main() arguments.
+ *     Returns: role as int, or 0 if argv[1] is missing or invalid.
+ *
+ *   hyades_handler_t
+ *     Type: void (*)(void). Each entry in the jump table must have this type.
+ *
+ * -----------------------------------------------------------------------------
+ * EXAMPLE: pingpong
+ * -----------------------------------------------------------------------------
+ *
+ *   Two processes run the same binary; each gets a role via argv[1]. Role
+ *   selects the order of actions (ping then pong, or pong then ping). The
+ *   run loop is entirely inside hyades_run().
+ *
+ *   Build (from the tests/ directory, so hyades.h is found):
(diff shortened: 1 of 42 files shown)


src/sst/elements/carcosa/injectors/faultInjectorMemH.cc
```diff
@@ -0,0 +1,315 @@
...
+        getSimulationOutput().output("Warning: Could not match '%s' to a MemHierarchy::Command, defaulting to Write\n",
+                                     install_cmd.c_str());
+        cmd = MemHierarchy::Command::Write;
+    }
+
+    // Initialize random number generator
+    generator = std::default_random_engine(rd());
+
+    // Parse injection probability
+    injectionProbability_ = params.find<double>("injectionProbability", 0.5);
```
mgoldstein322: This parameter has been removed from the base class.

src/sst/elements/carcosa/injectors/faultInjectorMemH.h
```diff
@@ -0,0 +1,149 @@
...
+#include "sst/elements/carcosa/injectors/faultInjectorBase.h"
+#include "sst/elements/carcosa/Components/PMDataRegistry.h"
+#include <map>
+#include <utility>
+#include <vector>
+
+namespace SST::Carcosa {
+
+// Enum to select basic fault injection logic or indicate a custom input
+enum injectorLogic {
```
mgoldstein322: Remove

src/sst/elements/carcosa/hyades.h
```diff
@@ -0,0 +1,158 @@
+/*
+ * Hyades - Vanadis control abstraction
```
mgoldstein322: I love the commitment to the naming :)

src/sst/elements/carcosa/PMDataRegistry.h
```diff
@@ -0,0 +1,316 @@
...
+// Government retains certain rights in this software.
+//
+// Copyright (c) 2009-2025, NTESS
+// All rights reserved.
+//
+// This file is part of the SST software package. For license
+// information, see the LICENSE file in the top level directory of the
+// distribution.
+
+#ifndef CARCOSA_PMDATAREGISTRY_H
```
mgoldstein322: Duplicate file?

mgoldstein322 (commented):
General comments:
- I'm not sure I really understand the PMDataRegistry. I can see the registry resolver maps a registry id to the actual registry; then the registry maps an event id to a port module command, but what are these commands? Is this how you instruct a PM to change its fault/swap out its injector?
- I know you said a decent amount of this was hardcoded to make the ping pong work, but I'm still pretty confused on the Hali code. My understanding is this: each `Hali` instance is a node in a ring of `Hali` instances. These nodes pass messages around the ring, but I'm not sure what the ring's purpose is. I've left various comments at my points of confusion in the `Hali.cc` file for more specific things, but I don't think I understand much beyond the MMIO coordination checking for specific commands at the magic address to see what to do.
- CarcosaMemCtrl: is this where most of Hali actually works? I see the custom event handlers and that they're open subcomponent slots, so I assume these are for the FaultInjEvents, HaliEvents, and CpuEvents. How does this differentiate between these events? Likewise, would CarcosaMemCtrl be loaded as the backend for a memory controller in SST?
- FaultInjectorMemH should be rewritten as a subclass of FaultInjectorBase. It contains a lot of old code from when I was just starting the Injector implementation that's either been completely changed or outright removed. I marked most of it, but probably not all of it.

nab880:
Hi Micheal, 

These are good suggestions and I update the pull request in a minute. 
In the meantime:

- I'm not sure I really understand the PMDataRegistry. I can see the registry resolver maps a registry id to the actual registry; then the registry maps an event id to a port module command, but what are these commands? Is this how you instruct a PM to change its fault/swap out its injector?
**PMDataRegistry contains "commands" that the PortModule has to interpret. The PortModule 'injector' interprets the commands and then modifies its behavior. Upon receiving a memory request the Injector will check the PMDataRegistry to see if a the FaultInjectionManager has a message corresponding to this MemEvent. **
- I know you said a decent amount of this was hardcoded to make the ping pong work, but I'm still pretty confused on the Hali code. My understanding is this: each Hali instance is a node in a ring of Hali instances. These nodes pass messages around the ring, but I'm not sure what the ring's purpose is. I've left various comments at my points of confusion in the Hali.cc file for more specific things, but I don't think I understand much beyond the MMIO coordination checking for specific commands at the magic address to see what to do.
**Hali is connected to Cores, so for every core there should be a corresponding Hali compnent. The Hali components are connected in a ring in the event that they need to broadcast information to each other. While we don't really have a use-case for this yet, I imagine it will be useful later on. i.e., we want to turn off all fault injection in a specific case.**
- CarcosaMemCtrl: is this where most of Hali actually works? I see the custom event handlers and that they're open subcomponent slots, so I assume these are for the FaultInjEvents, HaliEvents, and CpuEvents. How does this differentiate between these events? Likewise, would CarcosaMemCtrl be loaded as the backend for a memory controller in SST?
**CarcosaMemCtrl is largely just  an extension of the MemHierarchy MemoryController. I added additional functionality where we can exchange how the backend data is stored. The motivation for this was that if we want to reduce the overall memory requirements for SST but still want to have different pages of memory depending on the application. A Vanadis node might perform kernel A and then communicate with Hali it's done. Then Hali can change the contents of the memory backed before issuing the next kernel to be computed. **
- FaultInjectorMemH should be rewritten as a subclass of FaultInjectorBase. It contains a lot of old code from when I was just starting the Injector implementation that's either been completely changed or outright removed. I marked most of it, but probably not all of it.
**Yes, you are right this needs to be updated. **

bfeinberg (approved):
There is a lot here and I suspect we'll probably find some further pain points under major use, but tests look to cover the essential features so we're good to merge.

gvoskuilen (approved):
Approving based on @bfeinberg's review

PR #2638, opened 2026-03-04
Use SST_MPI_Comm_spawn_multiple in Airel PIN backend

### LINKED ISSUE #2624: Error "arielapi.c: MPI_Init called in arielapi.c but this file was compiled without MPI." although Ariel was compiled with MPI
In the latest Version of sst-elements (15.1.0) a binary that was compiled with the Ariel API throws the following error even if sst-elements was configured with `--enable-ariel-mpi`:

```
Error: arielapi.c: MPI_Init called in arielapi.c but this file was compiled without MPI.
```

## Details
 - sst-core was configured with `./configure --prefix=$SST_CORE_HOME --disable-mpi` (mpi needs to be disabled for Ariel to be able to run mpi binaries)
- sst-elements was configured with `./configure --prefix=$SST_ELEMENTS_HOME --with-sst-core=$SST_CORE_HOME --with-pin=$PIN_HOME --enable-ariel-mpi`. The output of the configure command confirmed that MPI for Ariel was enabled: `ariel_mpi : YES`.
- My binary that gets run by Ariel gets compiled with the following Makefile target:
  ```make
  mpi-test: mpi-test.c
          $(MPICC) $(CFLAGS) -Wl,-rpath,$(ARIEL_API_LIB_DIR) -o $@ $^ -I$(ARIEL_API_INCLUDE_DIR) -L$(ARIEL_API_LIB_DIR) -larielapi -fopenmp
  ```
- Ariel in my SST Configuration:
  ```py
  ariel = sst.Component("ariel0", "ariel.ariel")
  ariel.addParams( {
      "verbose" : "1",
      "maxcorequeue" : "256",
      "maxissuepercycle" : "2",
      "pipetimeout" : "0",
      "corecount" : str(corecount),
      "clock" : clock,
      "executable" : "./mpi-test",
      "arielmode" : "0",
      "arielinterceptcalls" : "0", # Do not intercept malloc/free 
      "launchparamcount" : 1,
      "launchparam0" : "-ifeellucky", # For Pin2.14 on newer hardware
      "mpimode" : 1,
      "mpiranks": 4,
      "mpitracerank" : 1,
  } )
  ```

## Possible Fix
I'm not familiar with Automake Makefiles, but here is how I was able to fix the issue: In `src/sst/elements/ariel/api/Makefile.am` I added `AM_CFLAGS = $(CFLAGS_LOCAL)`.

```diff
--- Makefile.am.old	2026-01-16 16:49:54.434124538 +0100
+++ Makefile.am	2026-01-16 16:50:03.922150464 +0100
@@ -1,21 +1,23 @@
 if SST_USE_ARIEL_MPI
 CC_LOCAL = $(ARIEL_MPICC)
 CFLAGS_LOCAL = -fopenmp $(ARIEL_MPI_CFLAGS) -DENABLE_ARIEL_MPI=1
 else
 CC_LOCAL = $(CC)
 CFLAGS_LOCAL =
 endif
 
 AM_CPPFLAGS += \
 	$(MPI_CPPFLAGS) \
 	-I$(top_srcdir)/src
 
+AM_CFLAGS = $(CFLAGS_LOCAL)
+
 lib_LTLIBRARIES = libarielapi.la
 
 libarielapi_la_SOURCES = \
     arielapi.c \
     arielapi.h
 
 libarielapi_la_LDFLAGS = -module -avoid-version -shared
 
 include_HEADERS = arielapi.h
```

plavin:
This PR updates Ariel to use `MPI_Comm_spawn_multiple` for launching applications when core is compiled with MPI support. This simplifies the build system as now the only parts of Ariel that depend on an MPI compiler are the test applications. This also makes launching applications more robust, as MPI applications are not supposed to call fork. Fixes #2624. 

This PR also adds new functionality to the Ariel API. Two new functions are added:
```
void ariel_output_stats_begin_region(const char *name);
void ariel_output_stats_end_region(const char *name);
```

These will each cause a message to be output to stdout, with the region name and the current simulation timestamp. This can be used to correlate stat dumps with locations in the source app. 

## Justification for changes to `fesimple.cc`
The file `ariel/frontend/pin3/fesimple.cc` required extensive changes. One tricky part about tracing MPI applications with PIN is that the MPI library will typically spawn its own threads when MPI_Init is called. This causes two issues: (1) the program has more threads than specified by the user in their Ariel config meaning the shared memory tunnel is not big enough, and (2) if MPI_Init is called before all of the application threads are launched, the MPI threads will receive lower IDs than the application threads. This makes ignoring them harder.

The first solution, which is removed by this PR, was to try and place an OMP parallel region before MPI_Init, so that the application threads would always be numbered 0..N-1. But this obviously only works for OpenMP programs and meant that the Ariel API needed to be compiled with `-fopenmp`. 

The new approach is to track which threads were MPI threads by checking if `libmpi.so` was in their callstack. This works but now we have to maintain a map of the thread IDs (typically [0,3,4,...,N+1] -> [0,...,N-1]). We then need to check this map when writing to the tunnel. PIN won't let us change how it numbers threads. In a future update, I hope to move this functionality to a class that wraps the tunnel so that the map can be queried in a single location instead of all over `fesimple.cc` .

## Known Issues
- There is a non-deterministic bug that will sometimes cause messages sent across the tunnel to appear in the wrong order to the arielcore. This will lead to errors such as `FATAL: ArielComponent[arielcore.cc:486:refillQueue] Error: Ariel did not understand command (128) provided during instruction queue refill.` or it may cause the program to hang indefinitely. This error seems to mostly affect `test_Ariel_test_ivb_pin`. 
- Calling MPI_Comm_spawn_multiple will cause an error if there are not enough slots in the allocation to run all the SST ranks and the application ranks in their own slots. To remedy this, we set `OMPI_MCA_rmaps_base_oversubscribe=1` in the MPI testsuite file. 
- The EPA backend needs to be updated to use SST_MPI_Comm_spawn_multiple instead of fork. Trying to launch an MPI app with the EPA backend will cause an error.
- The remap functionality in `fesimple.cc` may break if more application threads are launched than the the `corecount` parameter passed to the `ariel.ariel` component.


gvoskuilen:
This PR needs to be retargeted at the devel branch

src/sst/elements/ariel/frontend/pin3/pin3frontend.cc
```diff
@@ -204,27 +318,33 @@ int Pin3Frontend::forkChildProcess(const char* app, char** args,
...
         if(0 == app_env.size()) {
 #if defined(SST_COMPILE_MACOSX)
-        char *dyldpath = getenv("DYLD_LIBRARY_PATH");
 
-        if(dyldpath) {
-            setenv("PIN_APP_DYLD_LIBRARY_PATH", dyldpath, 1);
-            setenv("PIN_DYLD_RESTORE_REQUIRED", "t", 1);
-            unsetenv("DYLD_LIBRARY_PATH");
-        }
+            char *dyldpath = getenv("DYLD_LIBRARY_PATH");
```
gvoskuilen: This shouldn't be needed anymore since Pin doesn't run on mac. Could delete the whole section.

gvoskuilen (approved):
There's some extra code for Mac that isn't needed but it could get cleaned up later too.

src/sst/elements/ariel/frontend/pin3/fesimple.cc
```diff
@@ -415,30 +486,28 @@ VOID WriteStartInstructionMarker(UINT32 thr, ADDRINT ip, UINT32 instClass, UINT3
...
             ADDRINT* writeAddr, UINT32 writeSize, ADDRINT ip, UINT32 instClass,
-            UINT32 simdOpWidth, BOOL first, BOOL last )
+            UINT32 simdOpWidth )
 {
 
     if(enable_output) {
-        if(thr < core_count) {
-            if (first)
-                WriteStartInstructionMarker( thr, ip, instClass, simdOpWidth);
+        if(remap_id[thr] < core_count) {
```
gvoskuilen, line 506: missing `if (first)`

src/sst/elements/ariel/frontend/pin3/fesimple.cc
```diff
@@ -559,49 +628,33 @@ VOID InstrumentInstruction(INS ins, VOID *v)
     }
 
     UINT32 operands = INS_MemoryOperandCount(ins);
-    if (INS_HasScatteredMemoryAccess(ins))
-        operands = 0;
     for (UINT32 op = 0; op < operands; op++) {
         BOOL first = (op == 0);
         BOOL last = (op == (operands - 1));
 
-        if (INS_MemoryOperandIsRead(ins, op) && INS_MemoryOperandIsWritten(ins, op)) {
```
gvoskuilen, line 568: This `if` case needs to be added back in

gvoskuilen (changes requested):
Also marking WIP until we can figure out the frequent test failures.

PR #2646, opened 2026-03-18
Add missing setCpumask/setThreadId definitions and reassignCores virt…

nab880:
Updated to include some missing function definitions. 
Created virtual function virtual void reassignCores(Thread* thr) { } that previously existed in mercury.
While I think we could largely just remove the reassign core functinality, I included it as it was a part of sst-macro. We can comment this out if it's unneeded.


jpkenny (approved):
LGTM

PR #2656, opened 2026-04-18
mercury libfabric/sumi and mv2 compatibility fixes

nab880:
This PR wires Iris so Mercury-hosted simulations can run MVAPICH2 with the OFI netmod against the in-tree libfabric stack: register the SUMI provider, ship a PMI-1 (empty and untested PMI-2 support as well) compatible libpmi, and harden the SUMI provider enough for real fi_getinfo / EP / AV / CQ usage from MV2. Small Mercury tweaks keep C translation units and threading helpers consistent with that path.

### DIFF
--- a/src/sst/elements/iris/pmi/libpmi.cc
+++ b/src/sst/elements/iris/pmi/libpmi.cc
@@ -16,13 +16,13 @@ static SST::Iris::sumi::Transport* sstmac_pmi()
   return t->getLibrary<SST::Iris::sumi::Transport>("libfabric");
 }
 
-extern "C" int PMI_Get_rank(int *ret)
+int PMI_Get_rank(int *ret)
 {
   *ret = sstmac_pmi()->rank();
   return PMI_SUCCESS;
 }
 
-extern "C" int PMI_Get_size(int* ret)
+int PMI_Get_size(int* ret)
 {
   *ret = sstmac_pmi()->nproc();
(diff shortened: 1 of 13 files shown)


src/sst/elements/iris/pmi/libpmi.cc
```diff
@@ -0,0 +1,368 @@
...
+#include <cstring>
+#include <unordered_map>
+
+static SST::Iris::sumi::Transport* sstmac_pmi()
+{
+  SST::Hg::Thread* t = SST::Hg::OperatingSystem::currentThread();
+  return t->getLibrary<SST::Iris::sumi::Transport>("libfabric");
+}
+
+extern "C" int PMI_Get_rank(int *ret)
```
jmlapre: I don't think you need the extern "C" in front of all of these function definitions.  The compiler should pick it up properly from the declaration in the pmi.h header.

jpkenny (commented):
Justin's comments seem correct. I didn't find anything else obviously questionable. 

Installing pmi.h is an interesting corner case since we want headers and library to install to different paths. I don't have a better work around.

src/sst/elements/iris/pmi/libpmi.cc
nab880: You're correct, I removed them.

jpkenny (approved):
LGTM

PR #2665, opened 2026-04-28
Filter EmptyRankSync warning in tests

skuntz:
Filters out the EmptyRankSync warning in shogun and simpleComponents tests.


feldergast:
Let's fix this in core by not printing the warning unless checkpointing and/or interactive console is enabled.

PR #2667, opened 2026-04-28
General compiler warning fixes. Mostly missing overrides and printf s…

gvoskuilen:
Mostly missing overrides and printf specifier mismatches.

These were located using llvm 18.

### DIFF
--- a/src/sst/elements/osseous/rtlcomponent.h
+++ b/src/sst/elements/osseous/rtlcomponent.h
@@ -42,7 +42,7 @@ class Rtlmodel : public SST::Component {
 
 public:
 	Rtlmodel( SST::ComponentId_t id, SST::Params& params );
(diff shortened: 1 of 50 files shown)


src/sst/elements/osseous/rtlcomponent.h
```diff
@@ -42,11 +42,11 @@ class Rtlmodel : public SST::Component {
 
 public:
 	Rtlmodel( SST::ComponentId_t id, SST::Params& params );
-    ~Rtlmodel();
+        ~Rtlmodel();
```
feldergast: Why is this indented?

gvoskuilen: Fixed

feldergast (approved):

PR #2669, opened 2026-04-29
Fix GCC 15 and Apple Clang 21 compiler warnings

berquist:


berquist:
There are additional minor warnings in llyr on macOS, but it isn't being released for 16.0.0, so I skipped it for now.

berquist:
Eh, I'm almost done with llyr anyway...

feldergast (approved):

PR #2689, opened 2026-06-09
Fix sumi straddr buffer

nab880:
Fix a set of memory-safety and API-correctness bugs in sumi_av_straddr — the function that turns a SUMI/libfabric address into a human-readable string.

This caused a segfault (overflow the string buffer) on Ubuntu-24 builds of hgcc.


jpkenny (approved):
lgtm

PR #2691, opened 2026-06-24
merlin: write polarfly/polarstar topology files to a writable temp dir

nab880:
The polarfly/polarstar python modules wrote their adjacency list into os.getcwd()/<topo>_data/ and the C++ components read it back via getcwd(). When the working directory isn't writable (e.g. some CI runners) the python save() raised and model construction aborted right after the banner, leaving empty output. Generate the data in a tempfile.mkdtemp() directory instead and pass the path to the C++ side via a new 'data_path' param, removing the dependency on a writable cwd. C++ now fatals clearly if the file is missing.


src/sst/elements/merlin/topology/polarstar.cc
```diff
@@ -38,6 +38,8 @@ topo_polarstar::topo_polarstar(ComponentId_t cid, Params& params, int num_ports,
 
     total_endnodes      = params.find<int>("total_endnodes");
 
+    data_path           = params.find<std::string>("data_path");
```
feldergast, line 41: This should probably just use getOutputDirectory() (function inherited from BaseComponent) instead of passing in a path.

feldergast: Actually, this probably won't work because you can't currently get the final output directory path until the Filesystem object initializes.  And even if that is early enough, there's no way to get that directory in the Python (maybe we should add this to the API).  For now what you have is probably the best we can do.

src/sst/elements/merlin/topology/pymerlin-topo-polarfly.py
```diff
@@ -229,7 +230,7 @@ def __init__(self,q = -1, N = -1):
         self._declareClassVariables(["link_latency","host_link_latency","global_link_map","bundleEndpoints"])
         self._declareParams("main",["topo","q","hosts_per_router","network_radix","total_radix","total_routers",
                                     "total_endnodes","edge","name","algorithm","adaptive_threshold","global_routes","config_failed_links",
-                                    "failed_links", "GF", "vec_len"])
+                                    "failed_links", "GF", "vec_len", "data_path"])
```
feldergast: Do you want the data_path to be settable by the user? Adding it here will make it user visible. If you don't want it user visible, you can add it as a class variable above and then manually add that param to the component.

nab880: Good call — made it internal

src/sst/elements/merlin/topology/polarstar.cc
nab880: Exposing the output dir to the Python API would be a nice future improvement.

nab880:
Found another issue when running with the networkx module installed. The polarfly/polarstar topology builders check if networkx is present; if it is, they add print messages that aren't in the reference file, which breaks testing.

My current fix is to gate the print statements behind a higher verbosity level. The alternative is we could regenerate the reference files, but then things will break if we ever run without the networkx module

feldergast (approved):
LGTM

PR #2692, opened 2026-06-24
Add script that can be used to easily modify which elements build

gvoskuilen:
Adds a script that adds/removes .ignore files and reruns autogen.
The list of libraries can be specified either independently (e.g., 'merlin ember') or as a group ('network').
Dependent/commonly used libraries for the requested ones are automatically added.

The script assumes it is being run in the top-level sst-elements directory.

### DIFF
--- a/select_build.py
+++ b/select_build.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+
+import sys
+import os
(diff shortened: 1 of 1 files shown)


feldergast (approved):
LGTM

PR #2697, opened 2026-07-04
mercury/nic: key message reassembly by (source, flowId)

nab880:
This fixes an edge case bug where the completionQueue would interleave messages if there is a flowId collision.

flowId is only unique per sender. When there are many senders and one receiver collisions can occur. Different sources packets get reassembled in the same slot.

The fix was to add a recvCqFlowKey that packs source into the top 16 bits and flowId into the lower 48 bits to avoid collisions.


jpkenny (commented):
I'm just wondering if we should just send the src id to RecvCQ::recv() and not risk the source id overflowing the 16-bit int at some point. It would be slightly less efficient for the simulation but wouldn't affect the simulation results.

nab880:
Joe's comment should be resolved.

jpkenny (approved):
Thanks for the changes, looks good now.

PR #2698, opened 2026-07-07
Use traceEnabled on all trace types in Prospero

plavin:
It seems that the `-d` flag in the Prospero trace tool only works for binary traces. This flag disables tracing at the start of programs until `prospero_enable_tracing` is called in the API. This commit ensures it is respected for text traces as well as binary traces. 

I fixed some whitespace issues which unfortunately make the diffs look bad.


gvoskuilen (approved):

gvoskuilen:
@plavin Can you merge the latest devel into this branch? It is failing the pre-commit check and the latest devel should fix that.

plavin:
@gvoskuilen Done

gvoskuilen (approved):

PR #2706, opened 2026-07-14
Change all #!/usr/bin/env python to python3

### LINKED ISSUE #2705: Prefer python3 vs. unversioned python
According to [PEP 394](https://peps.python.org/pep-0394/) we should prefer `#!/usr/bin/env python3` over any other incantation.

jmlapre:
All Python files should start with `#!/usr/bin/env python3`.

Closes #2705


gvoskuilen (approved):
LGTM

PR #2713, opened 2026-07-24
Anytopo test reference and packaging updates

berquist:
This is a follow-up to https://github.com/sstsimulator/sst-elements/pull/2598.

The Anytopo tests weren't running because we didn't (and still don't) have the correct Python libraries installed on our testing infrastructure.  In an effort to do that (https://github.com/sstsimulator/sst-sqe/pull/1245) we discovered two things:

- One of the test references isn't correct (tested on the cluster head node)
- Testing on the result of `make dist` fails because test files were missing from `Makefile.am`.

Once this PR is in, the above SQE PR can be merged.

### DIFF
--- a/src/sst/elements/merlin/tests/refFiles/test_merlin_anytopo_jellyfish_test.out
+++ b/src/sst/elements/merlin/tests/refFiles/test_merlin_anytopo_jellyfish_test.out
@@ -1,6 +1,6 @@
 Loaded NetworkX graph with 16 vertices and 32 edges
   Offered   Average
     Load    Latency
-     0.50      268.375 ns
+     0.50      262.401 ns
 
 Simulation is complete, simulated time: 400 us
--- a/src/sst/elements/merlin/Makefile.am
+++ b/src/sst/elements/merlin/Makefile.am
@@ -94,6 +94,12 @@ libmerlin_la_SOURCES = \
 	pymerlin-router.py \
 	interfaces/pymerlin-interface.py \
 	target_generator/pymerlin-targetgen.py \
+	topology/anytopo_utility/DallyDragonfly.py \
+	topology/anytopo_utility/HPC_topos.py \
+	topology/anytopo_utility/__init__.py \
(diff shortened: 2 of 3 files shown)


src/sst/elements/merlin/tests/refFiles/test_merlin_anytopo_jellyfish_test.out
```diff
@@ -1,6 +1,6 @@
 Loaded NetworkX graph with 16 vertices and 32 edges
   Offered   Average
     Load    Latency
-     0.50      268.375 ns
+     0.50      262.401 ns
```
jmlapre, line 4: I'm confused why this test failed: https://sst-jenkins.sandia.gov/job/SST__Autotest_OSX-15-XC16_OMPI-4.1.6_PY3.10_sst-elements/1320/consoleFull

It looks like it's complaining but the change is exactly what you made here???

berquist: The SHA of the commit being used is correct.  It shouldn't be from testing the merge with `devel`, since that's what running locally also defaults to.  I'll try running locally on that machine, without the SQE changes in https://github.com/sstsimulator/sst-sqe/pull/1245, and also look at the Jenkins config for this job.

I am surprised that the test is being executed at all, which means the Python dependencies are present on that machine (not sure how).

berquist: To the last point, I am seeing failures on nightlies for a few of these tests now, for tests that pass on the head node.

I am not sure how we normally handle (floating point) differences in required reference data for different machines, I'll need to look at the testing framework.

berquist:
I can't reproduce the problem locally on that machine, using bamboo, but it keeps happening in Jenkins.  I've uninstalled the stray Python packages needed to even run these tests from our testing environments, so I'm going to punt this problem down the road.

PR #2714, opened 2026-07-28
iris/mask-mpi collective registry

nab880:
Add an ELI-backed registry for selecting Sumi collective algorithms and move
collective configuration under the `collective.*` namespace.

This PR also separates all-reduce topology policy from algorithm selection.
Selecting an all-reduce algorithm no longer implicitly disables Mercury's SMP
hierarchy

Algorithms are selected with namespaced parameters


jpkenny (approved):
LGTM

PR #2716, opened 2026-08-04
Remove dead variable.

tdysart:
The removed variable "id" is not set in the typical constructor but then used in setup() - this was causing issues with our internal testing of mordred.  Since this variable appears to have been replaced by "net_id" we have removed it from the nic.


jleidel:
This bug was found in experiments with the `mordred` subcomponent: 
https://github.com/tactcomplabs/mordred/tree/debug_const

feldergast (approved):

PR #2722, opened 2026-09-02
Mercury: Preserve ARM64 FP state across Mercury contexts

nab880:
Fix Mercury’s ARM64 fcontext implementation to preserve the AAPCS64 callee-saved floating-point registers d8–d15 across context switches. Updates both ELF and Mach-O assembly.


jpkenny (commented):
All looks fine to me, I just want to document how this differs from  historical practice in this code base:

To date, we've never needed to change the assembly here, which all comes straight from boost fcontext. We only modified the interface slightly to hook it into macro/mercury. How were these asm modifications actually obtained, and how would we go about preserving their effect if we need to bring in other boost fcontext updates down the road?

nab880:
The bug existed in the boost library from October 2015 until November 2016 and Our copy sits in that window. sst-macro picked the fix up at some point (looks like Dec 2021); Mercury's copy never did.

So if we shouldn't really need to preserve this change, every upstream revision since  boost 1.63 has it. The Mercury-local delta is still just the symbol prefix. I could update to a current Boost tag with as a separate PR as well, if that's preferable.

jpkenny:
> The bug existed in the boost library from October 2015 until November 2016 and Our copy sits in that window. sst-macro picked the fix up at some point (looks like Dec 2021); Mercury's copy never did.
> 
> So if we shouldn't really need to preserve this change, every upstream revision since boost 1.63 has it. The Mercury-local delta is still just the symbol prefix. I could update to a current Boost tag with as a separate PR as well, if that's preferable.

Ah, ok, so this is just an update to get back aligned with more recent boost. That's fine then. Preplexing if it was fixed in macro in 2021 as Mercury was split off long after that.

jpkenny (approved):
LGTM

